PR #1234, opened 2026-08-24
fix: repair seven defects found in the installed rc.10 sweep

wlsdks:
## Summary

Walking the installed rc.10 app surface by surface turned up seven defects that unit tests could not see, because each one only appears in a real session. Each fix carries a regression test.

**Keyboard shortcuts were dead for this product's primary audience.** Combos matched the character a key *produced*, so with a Korean input source the D key emits a jamo and never equals `"d"`. The docs drawer, search, open-folder and docs palette were all unreachable while the shortcut sheet advertised them. Matching `event.code` makes a letter shortcut mean the key's position, which is what it always meant. IME composition keystrokes are left alone.

**The relation editor could wedge permanently.** It stays mounted for its exit animation and is keyed by node id, so reopening the same node returned the previous instance — and the success path never reset `saving`. The panel came back showing an already-written change with every control disabled, including "edit again". Measured frozen past 90s on the shipped build.

The rest are smaller but visible in the same walk:

- Hangul fell through to a system monospace face and rendered with broken spacing wherever a label used `font-mono`
- the relation legend drew depends-on and related-to at a size where their difference could not be seen (the canvas encoding is unchanged; only the swatch grew)
- the notification bell looked identical whether or not anything was unread
- the INDEX panel covered the first toolbar chip
- the git location read as loose text beside three bordered buttons
- a running agent turn showed a motionless dot

## Test plan

- `pnpm checks:changed -- --run` — all recommended checks
- `pnpm test:contracts` — 2047 passing, lint-warning ratchet at 0
- `pnpm source:language` — 0 violations
- `pnpm exec tsc --noEmit`, `pnpm knip`, `pnpm docs:surface:check`, `pnpm vault:validate`
- Playwright `overflow-sweep` and `ontology-ui` pass
- Verified in the built static bundle: Korean `ㅏ`+`KeyK` opens the palette and `ㅇ`+`KeyD` opens the docs drawer; both were dead before
- The meaning-editor test was probed in both directions — RED without the reset, GREEN with it

### Known pre-existing failure

`tests/e2e/map-viewport-reframe.spec.ts` fails on this machine (0.359 vs 0.35 threshold). Verified it fails identically on a **clean tree with zero changes** — machine-speed sensitive, not caused by this PR.

https://claude.ai/code/session_01W194qVZSimwSpLaeQs9xNs

### DIFF
--- a/app/globals.css
+++ b/app/globals.css
@@ -358,7 +358,15 @@
 
   /* Font family */
   --font-sans: var(--font-pretendard), system-ui, sans-serif;
-  --font-mono: var(--font-jetbrains), ui-monospace, monospace;
+  /* JetBrains Mono carries no Hangul, so before Pretendard was listed here every
+     Korean string set in `font-mono` fell through to the system monospace face —
+     which renders Hangul full-width and substitutes fullwidth punctuation. On the
+     shipped build that turned eyebrow and data labels into visibly broken spacing:
+     wide gaps between syllables and fullwidth parentheses. Latin and
+     digits still resolve to JetBrains first, so slugs, paths and tabular numerals
+     are unchanged; only Hangul now lands on the Korean face the rest of the app
+     already uses. Guard: tests/contract/mono-hangul-fallback.contract.test.ts */
+  --font-mono: var(--font-jetbrains), var(--font-pretendard), ui-monospace, monospace;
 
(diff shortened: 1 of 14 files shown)


wlsdks:
Superseded by #1240. These commits are carried there, rebased onto current `main`.

PR #1235, opened 2026-08-24
feat: hold the write checkpoint in the server that performs the write

wlsdks:
> Stacked on #1234. Review that first; base retargets to `main` when it merges.

## Summary

Decision **(113)** in `docs/DECISIONS.md` records the measurement and the reasoning; this is the implementation.

### Why the previous gate could not have worked

The gate an agent shows is the *agent's* gate, and it does not necessarily cover this server. Codex's approval policy says so itself — it decides when approval is required *"before executing a **command**"*, and MCP tool calls were never in that scope. Reading `codex-acp` confirms the shape: the adapter builds permission requests from `item/commandExecution/requestApproval` and `item/fileChange/requestApproval`, with no MCP-tool equivalent.

That is why decision (111) measured a `read-only` Codex session changing a vault with no card, and why no third-party configuration could have fixed it.

### The lever we already own

The same adapter forwards `mcpServer/elicitation/request` into ACP `session/request_permission`, and this server's SDK (`@modelcontextprotocol/server` 2.0.0, latest published) implements `elicitation/create`. So the checkpoint goes **where the write happens** — it reaches every client that ever connects, not just the one that was measured.

### One session, one checkpoint

Held by whoever can actually hold it. A runtime with app-owned config isolation (Claude) keeps its own gate and the server gate stays **off**, so nobody is asked twice. Everything else — including a runtime nobody has measured — gets the server gate.

**Fail closed:** if the gate is on and the client never declared the `elicitation` capability, the write is *refused*, not performed.

> Isolation is not gate ownership. Rust's `ISOLATION` says the app can control a runtime's config directory; owning the gate is the stronger claim that the configuration actually asks *for MCP writes*. Conflating the two is how (111) happened. `runtime-gate.test.ts` holds the one-directional invariant.

## Test plan

Wire evidence on disposable vaults:

| gate | client | asked | vault |
|---|---|---|---|
| off | no elicitation | no | **wrote** (7 → 8) |
| on | no elicitation | no | refused (7 → 7) |
| on | accepts | yes | wrote (7 → 8) |
| on | declines | yes | refused (7 → 7) |
| on | **real Codex** | yes | **refused** (7 → 7) |

The Codex row ran `codex exec` against an isolated `CODEX_HOME`. The same `add_concept` call that (111) measured writing unasked returned *"The change was not approved (decline). No change was made."* and left the vault untouched.

- `mcp/src/write-consent.test.mjs` — 10 passing
- `src/features/acp-session/model/runtime-gate.test.ts` — 8 passing, includes the Rust↔TS drift invariant
- `pnpm test:contracts` 2047 · `integration:mcp:write` 45 · `dogfood:verify` · `docs:surface:check` · `tsc` · `knip` · `source:language`
- **No new dependencies.** Repo and SDK are both MIT.

## Not in this change

**Codex in-app chat is not restored.** Decision (111) stays active. Half of its condition is now measured; the direct-write half has not been re-measured with both gates live, and (111) says explicitly not to restore from a direct-file sandbox result. That needs the Rust isolation entry, a rebuilt app, and installed-app acceptance — a separate round.

https://claude.ai/code/session_01W194qVZSimwSpLaeQs9xNs


wlsdks:
Superseded by #1240. These commits are carried there, rebased onto current `main` — which had moved past this branch's base by four merges.

PR #1276, opened 2026-08-26
chore: point /download at v1.0.0-rc.14

wlsdks:
Regenerated by `pnpm desktop:release-run` after v1.0.0-rc.14 published. Byte sizes and SHA-256 values are read from the published release, not typed.

Without this the download page keeps advertising the previous build.


wlsdks:
Superseded by #1300, which points the download surface at v1.0.0-rc.17.

PR #1335, opened 2026-08-31
docs: read the 24 saved answers and record what a grader can see

wlsdks:
> Stacked on #1334. GitHub will retarget this to `main` when that merges.

## Why

#1334 showed the machine score could not compare the two sides, and that only a person reading the answers could. This takes the first step down that route and builds the tool the real grading will use.

## What it is — and is not

**A model-graded screening pass, not a verdict.** It was produced by the same assistant that built the scoring split, which is weaker than the human grading `rubric.md` asks for and does not replace it. Two limits are stated in the document rather than hidden:

- the grader is not a person;
- an answer written with a vault names its concepts and one written without does not, so the sides stay recognisable from wording. Shuffling prevents grading in a convenient order; it does not blind anyone.

**None of these numbers goes in the README.** The public claim stays where #1334 left it.

## New tool

```bash
pnpm benchmark:blind-set --run-id=2026-08-31-gb-r3
```

Writes a packet where each answer carries an opaque id in a shuffled order, plus a key mapping them back. The shuffle is seeded from the run id, so two graders get the same packet in the same order and their grades line up cell by cell. Regenerating the committed packet is byte-identical.

## What the reading found

Averages are close and are the least interesting part (correctness 0–3, 12 answers per side): 2.50 without Atlas, 2.83 with. The four questions do not behave alike:

| Question | What it asked | Without Atlas | With Atlas |
|---|---|---:|---:|
| G1 | who owns a new discount rule, what to read first | **3.00** | 2.67 |
| G2 | should reconciliation move into checkout, what is the recorded relation | 2.00 | **3.00** |
| B1 | what an acknowledgement change touches, what is only declared | 2.33 | **2.67** |
| B2 | where permission evaluation belongs, what to verify next | 2.67 | **3.00** |

- **Atlas won where the question was about something recorded.** On G2 both sides correctly said *no* — but only the Atlas side could quote `checkout depends_on inventory-sync` and its stored reason. The control reconstructed the same idea from product prose, which is not a recorded relation. On B2 the Atlas answers stated the boundary in **both** directions; no control answer did.
- **Atlas lost the orientation question, and the reason is actionable.** Both sides correctly said Purchase/Checkout owns a discount rule. They differed on the second file: the control went to the cart summary (where a discount actually shows up), the Atlas side followed the `depends_on` edge to inventory-sync. **A declared dependency became a reading order and pointed away from the change.** One Atlas answer avoided it by labelling cart-summary "source-discovered, not an Atlas-declared anchor" — the right instinct, in 1 of 3.
- **The only unsupported claim in all 24 answers came from the side without a vault** — it borrowed Access Control's permission verbs ("read", "coordinate") and reused them as Coordination capability names.

## Caveat that limits all of it

The fixture's source is four one-line files; the meaning lives in the subject's README and product doc, which **both** sides read. So this compares a vault against good product prose, not against code — the same condition `FINDINGS-2026-08-25.md` named on this repository, still unresolved.

## Test plan

- `pnpm checks:changed -- --run` — 10 checks passed
- `pnpm test:benchmark` — 19/19 (4 new)
- `pnpm exec eslint scripts/benchmark-blind-set*.mjs` — clean
- Packet regeneration is byte-identical (md5 verified)

## Next

A person grades the same packet; where the two gradings disagree is where the criteria are unclear. Separately, the G1 reading-order finding is worth treating as a product question: should a handoff present a declared dependency and a likely-touched file in one list, in one voice?

https://claude.ai/code/session_01NGDyqY1Hi1mA9QsouY7kv4


wlsdks:
Superseded by the PR above — this one was closed automatically when its base branch was deleted on the merge of #1334. Same commits, rebased onto `main`.

PR #1488, opened 2026-09-06
fix: read a Korean check report's counts, and say the wiki log line in the screen's language

wlsdks:
## Summary

Post-merge inspection of the installed build (main 7ac676131) on a Korean screen found one defect in #1486's wiki log: the localized Check-the-wiki brief asks for a Korean count line, `describeLintTurn` knew only the English labels, so `wiki/_log.md` recorded `ran; counts not stated` and the Library header showed that English sentence.

- The parser reads the counts under their English or Korean labels; the log keeps its machine form (`disagreement 0 · superseded 2 · …`) so `grep` reads the same words in every vault.
- The Library header translates that form back (`localizeWikiLogSummary`), including a compile's `(new)` / `(revised)`.
- The empty-wiki copy named `wiki/_template.md` in a folder that has none (the dogfood vault); it now says the template comes from Settings › Workspace › Start a wiki.

## Test plan

- `pnpm checks:changed -- --run` — 11 checks pass (lint, unit, contracts, source language, i18n messages, docs-vault).
- Installed app on the wiki-only and map-plus-wiki fixture vaults: Check the wiki ran end to end on both; candidates, Propose chip gating, session survival and the map all verified on merged main. The count line defect reproduced on both runs.


wlsdks:
Opened from a stale branch head; superseded by the PR from fix/wiki-log-ko-summary.

PR #1528, opened 2026-09-08
feat(library): show real file work on the graph

wlsdks:
## Library work, grounded in actual events

Implements the owner-selected first slice: retain the knowledge/provenance graph, add a separate live work projection and inspectable recent file observations. No fake progress, idle drift, renderer replacement, new route, or write-authority expansion.

- Exact structured ACP/local tool targets; real `read`/`edit`, provisional writes and permission waits are distinct.
- File-change receipts come from observed wiki revisions or successful local write results, not a tool title or permission grant. Unknown targets do not light arbitrary nodes.
- Reserved activity lane avoids shifting the graph when work starts. Recent file controls work outside chat, by keyboard and on touch.
- Hidden reader/tab and reduced-motion handling; previous turns and other vaults cannot reappear as current work.
- Fix the pre-existing empty Sonner hotkey array that made every key steal focus while a notification was present.

## Verification

At commit `256245b1b`, all 33 recommendations from `PLAYWRIGHT_STATIC=1 PLAYWRIGHT_BASE_URL=http://127.0.0.1:3218 pnpm checks:changed -- --run` passed, including static build, desktop performance budget, contracts, types/lint/docs, unit regressions and the five static browser/accessibility cases. A separate existing graph regression spec passed all 5 cases (drag, hover, zoom/fit, idle-frame sleep and reduced motion). The surface census RED→GREEN probe also passed. The subsequent test-fixture-only typing cleanup (`db40cd081`) passed all 3 newly recommended checks, including all 3 activity browser tests and TypeScript. These results cover this branch's baseline, not an integration with the newer main.

Astra independently reviewed source twice; Terra implemented bounded fixes. Review caught and corrected `edit` omission, cross-operation clearing, previous-turn resurrection, hidden-frame work, folder baseline errors, non-actionable permission waits and dropped terminal reads (including batched completion). This is not a live-provider qualification: the native protocol is a deterministic fixture and no user vault was mutated.

## Deliberately not merge-ready

All 11 normal pre-push lanes passed at `db40cd081`, including full lint, unit, contract, typecheck and documentation checks. CI E2E remains separate.

- Main advanced during implementation: #1526 and #1527, with #1527 changing the Library into a graph-beside-reader layout. This branch is not yet integrated. A read-only merge preview reports conflicts in LibraryPage, changelog/generated docs and the surface census. Preserve the new main direction when integrating; the old hidden-reader proof must be rerouted for that layout.
- Final Computer Use accessibility capture is blocked: macOS reports visible Chrome/Atlas windows but no accessible window, although permission inventory says granted. Earlier baseline/checkpoint captures succeeded. The owner has been asked to unlock/re-enable Orca Computer Use. Do not bypass this proof.
- Real macOS read recording exists: 4 seconds, uniform30fps extraction, full-pixel comparison over the entire interval has zero stalls (cv0.50; one spike). Reduced motion preserves Reading with an identical canvas over1200ms. Final motion/design approval is deferred until it can be bound to the required fresh accessibility capture.
- Selected design council seats (lead, interaction, motion, infoviz) and the final structural ledger entry remain pending; no council approval is claimed.
- A first-page sync overlap of9.3px was independently reproduced in baseline530382f, with matching coordinates. It is not introduced by this PR, and general graph readability is not claimed solved.
- Existing ontology capability updated via MCP; validation and full compile have zero problems. Meaning finalization refused `source_receipt_unavailable`; no receipt was fabricated.

## Council utility (pending)

**Pre-review decision**: the owner's stable-graph plus real-event work layer.
**Selected seats / first positions**: lead, interaction, motion, infoviz /0 (required final evidence blocked).
**Rebuttal**: none.
**Decision delta**: pending; source correctness review strengthened proof, separately from council.
**Unique contribution**: none claimed for an unconvened council.
**Review footprint**: first0 · rebuttal0.

Local evidence packet: `/tmp/atlas-library-live-proof.9zT76r/REVIEW.md`; recordings/frames/metrics remain outside the repository. Main's unrelated dirty CodeGraph work is excluded.


wlsdks:
## Integrated with current main; every routed check passes

Picked this up because the branch had stalled and could no longer be merged. **Merged rather than rebased**, so the two original commits stay exactly as written — a rebase here would have meant a force-push over someone else's commits.

### The conflict that mattered

`#1527` put the Library's graph beside the page it is about, which made that pane a **row**. The work lane is a horizontal band (`h-28 flex-none`), so left as they were the lane would have stood up as a **third column** between the graph and the reader.

Resolved by moving `#1527`'s row one level in and letting the lane span the pane above it:

```
library-reader (flex-col)
├─ LibraryWorkActivityStrip      ← the lane, height reserved
└─ div (graphBesideReader ? flex-row : flex-col)   ← #1527's row, unchanged
   ├─ library-graph-column
   └─ library-document-column
```

The two alternatives both fail a real state: a lane inside the row *is* the third column, and a lane inside either column disappears with it — below `lg` the graph column is exactly what hides when a page opens, which is when work most needs to stay visible. Above the row it survives all four states, and it now stays visible while the graph stands beside an open page, which the branch's original placement could not do.

### Other resolutions

- **Surface ratchet → 41.** Both sides independently wrote 39 → 40. They are different surfaces (main's Mac download menu, this branch's work lane), so the real count is 41; both reasons are kept. Left at 40, one of the two surfaces would have gone uncounted with the gate still green.
- **Unreleased changelog** keeps main's wording and adds only this branch's fact, compressed to 899 of its 900 bytes.
- **Generated docs-vault output** rebuilt via `pnpm docs-vault:build`, never hand-merged.

### Verification, against integrated main rather than this branch's own baseline

- `pnpm test:run` — **978/978 files, 9527 tests, 0 failures**
- `pnpm exec tsc --noEmit` · `pnpm lint --max-warnings 0` — clean
- `pnpm checks:changed -- --run <50 changed paths>` — **all 33 routed checks passed**, including `build` + `desktop:perf`, `test:contracts`, `vault:validate`, `knip`, `docs:links`, `changelog:check`, `test:i18n:messages`
- `library-work-activity.spec.ts` **3/3** (desktop, rejection, touch) and `a11y-open-surfaces.spec.ts` **2/2**
- All 11 pre-push lanes passed

The touch case is the one that could have caught the relocation — it measures a 44px hit target and asserts zero horizontal overflow at 390×844. It passes.

### Still open — unchanged from the original description

- **Computer Use accessibility capture** is still not done. Not bypassed, not substituted.
- **Design council** (lead, interaction, motion, infoviz) remains unconvened, and the layout change above is structural, so it is exactly the kind of thing that routing exists for. Passing tests say nothing broke; they do not say the placement is right.
- The evidence packet cited in the description (`/tmp/atlas-library-live-proof.9zT76r/REVIEW.md`) is **gone** — `/tmp` did not survive. The verification above is freshly measured, not that packet re-read.

Kept as a draft.

https://claude.ai/code/session_01M3nGUAZHRQedWEZzL2D2ah

PR #1536, opened 2026-09-10
feat(map): give the walked constellation one fact per channel

wlsdks:
## Summary

Two design seats reviewed the walked-path constellation and measured the same encoding failing from both ends. Neither defect was visible from the source, and one of them was hidden *by* the source — a comment claiming the opposite of what the code did.

**design-infoviz** sampled the canvas back-buffer and found the engraved child count inside a walked node at **1.00:1 against its own face** — erased. `drawNodeStar` built its bloom as a radial gradient from `radius * 0.35` and filled it with a plain disc; a radial gradient paints everything inside its inner circle with stop 0, so the face took a solid additive wash at α 0.62. The function's header had read "the face is never filled" for three days. It is an annulus now, so the sentence is true by construction.

The same seat proved brightness cannot rank here at all. Additive light on a dark canvas clips, so the recency ramp's own adjacent-step ratio is 1.41 at two stops and 1.17 at four — under any discrimination threshold even with the twinkle at zero. Measured on screen, a seven-step walk read 5 > 6 > 7 > 3.

**design-motion** recorded the real display and found the destination missing: the focused node was deleted from the star set under a rule written when the mark was a shoe print *beside* the node and kept after the mark became the node. The stop a person had just arrived at drew 19× darker than the ones behind it. The same recording caught an interruption flash and a distance-rule violation.

Both seats independently measured the detail panel beating the constellation for attention (5.3:1; 66.9% of the first frame after the trigger).

## What changed

| | before | after |
|---|---|---|
| numeral vs its own face, walked node | **1.00:1** (erased) | **5.30–5.55:1** |
| newest stop | deleted from the walk | lit, in the selection's indigo, numbered |
| brightness | claimed order, delivered 5 > 6 > 7 > 3 | says *walked*, and nothing else |
| order | brightness | the ordinal and the popover list |
| direction, reduced motion | nothing at all | a static chevron on every walked relation |
| travelling light | 3 lights, 59 / 148 / 169 px·s⁻¹ | 1 light, one speed, in walk order |
| close mid-sweep | **+47%** flare in one frame | fades from where the sweep reached |
| detail panel under the lens | wins the screen | recedes to 0.46, restored on hover/focus |
| settings preview | the glyph the owner rejected | the mark the map draws |

The walked star also lost its diffraction cross. design-infoviz offered two ways to stop it colliding with the magnitude spike — separate them by form, or drop it. Form went in first, turned 45° and clamped clear of the label ring, and rendered as an X drawn across the node. Dropping it removes the overlap instead of arbitrating it, and costs nothing: the cross said "here is the end of the walk", which the indigo star and the ordinal already say. The magnitude spike stops standing down on a walked node and gets its fact back.

Retiring the preview's use of the glyph painters made them dead code, and the ratchet said so. Following that down, six settings controls — fill, outline weight, bloom, along-the-line repetition, its density and its side — were read by nothing but the code that had just died. They are gone; a preference saved while they existed still loads, and a test states that promise.

## Corrections to the record

- Commit `67708d757`'s message claimed "both new assertions were probed red". Only one was; the other matched a doc comment and could not fail. Corrected in `bd5e32f5a` and the gate rewritten.
- Three comments asserted things that were false when measured: that the star's face was never filled, that the stars hold still because brightness means recency, and that the four-second glint lap was licensed by the 2026-09-08 drift removal — a decision that governs the Library's canvas, not this one.
- The composite-licence gate then went red against the comment explaining the very removal it was checking — the second time in one day a gate here read prose. It strips comments now.

## Test plan

- `pnpm checks:changed -- --run` — **18 passed** (tsc, lint, knip, unit, contract, e2e overflow sweep, i18n, decisions, docs-vault).
- Rendered evidence via Chrome DevTools on the live dev server at 1376×900: the map with the lens open and a four-stop walk, and the settings panel.
- Canvas back-buffer sampling (`getImageData`, dpr 2) for every number in the table above.
- **Gate probes, both directions.** The interruption fix: reverting only the clock line reproduces **+47% in one frame at 349 ms** against a close at 300 ms; restored, the largest rise near the close is its own in-flight ignition. The composite-licence gate: red on a restored spike call, red on a deleted composite restore, green on the source.

## One unrelated fix, carried because it blocked this branch

The transient-surface sweep went red on `gateway-hero-cta → menu` does not close on Escape. It reproduces on **clean `origin/main`**, so this branch surfaced the defect rather than causing it — main's own run skipped that spec through the e2e impact plan. It is fixed here, in its own commit, because it blocks the merge.

The report understated it: the menu had no keyboard at all. Focus was moved to the first row by a `requestAnimationFrame` scheduled the moment `open` flipped, but `Surface` gates on `usePanelPresence`, which sets `mounted` inside an effect — so on that render the surface returns `null` and the ref the frame callback reads is still null. Measured on the built export, `document.activeElement` was still the trigger button **420 ms** after the click. With no row focused, Escape, ArrowDown, ArrowUp, Home and End all landed on the trigger, whose own handler answers only ArrowDown. The sweep caught the Escape half; the arrow keys were checked by nothing.

A callback ref fires when React attaches the node, so there is no frame to lose the race in. Escape also moves to the document, because whether a menu closes must not depend on where focus happened to land. Five of the six new unit cases were probed red against the restored defect and green against the fix.

## One CI failure that was genuinely flake

`Playwright (chromium 2/3)` failed one run on `download-hero-echo.spec.ts` — 3 of 3 attempts. It is unrelated to this branch: nothing here touches `src/views/download`, and the hero's headline copy is unchanged. The test polls a CSS typing animation against a hard 4-second deadline and reached 51 of 59 characters, so it needs about 4.6s on a loaded runner. Re-running that job alone on the identical commit passed 11/11. Marginal budget, not a regression — but it is real fragility in that spec and worth its own fix.

## Deferred

`/motion-verify`'s real-display recording is **not** part of this proof. Two takes were captured and both came back as the machine's screensaver; `caffeinate -u` could not dismiss it, so the session is locked. Per the skill's own precondition that is a deferred proof, not a passed one — the in-browser rAF sampling above diagnoses the motion but cannot approve it. design-motion's recording covers the pre-fix build only.

https://claude.ai/code/session_014w6Pn2ktVCthCd6DcdWcPo


wlsdks:
## `/motion-verify` — the deferred real-display proof, now captured

This PR shipped with the real-screen recording marked **deferred**: two takes came back as the machine's screensaver and `caffeinate -u` could not dismiss it, so the session was locked. It is unlocked now and the recording exists. Filed here rather than in a new PR because it is this PR's owed evidence.

**Capture** — `screencapture -v -D 1`, built-in display 3456×2234, 9.30 s, extracted at a uniform 30 fps (279 frames) with `ffmpeg`. Chrome on the dev server at merged `main` (`67f4f5ca2`), `/ko/topology/?guides=off&e2e=1`, a four-stop walk (고객지원 → 배송 → 주문 → 결제), trail lens opened by clicking its own chip. Crop `(1150, 560)–(2620, 1820)`, chosen from a phase strip before any statistic was computed.

**Phase strip** (6 frames, 165 ms apart, from the click):

| t | what is on screen |
|---|---|
| 0 ms | lens closed, nothing lit, detail panel at full strength |
| 165 ms | 고객지원 (1) coming up; panel already receding |
| 330 ms | 고객지원 lit, the walked line to 배송 appearing |
| 495 ms | more lines lit, 결제 starting |
| 660 ms | 주문 (3) igniting, 결제 lit in indigo |
| 825 ms | all four settled |

The walk replays in the order it was walked, and the panel demotion is visible across the sequence.

**Adjacent-frame change**

| interval | n | mean | cv | min | stalls | spikes |
|---|---|---|---|---|---|---|
| ignition sweep, f099–f126 (933 ms of motion) | 27 | 0.3491 | 0.85 | 0.0386 | **1** | 2 |
| settled, f130–f269 (4.6 s) | 139 | 0.0108 | 0.62 | 0.0016 | 4 | 4 |

**On the one stall — it is a lull, not a freeze.** f123 measures 0.0386 against a 0.0929 threshold, but the *true* resting floor on this crop is 0.0004–0.0017 (frames f095–f098, before the click registers): f123 is **102× the resting floor**, so the canvas was still changing at a hundred times its idle rate. It is not the signature the check is looking for — idle-gate sleep or a dropped browser frame both sit at the floor, not two orders of magnitude above it.

My first pass reported 3 stalls by starting the interval at the click rather than at the first moving frame, which counted three resting frames as stalls — the skill warns about exactly that ("expected resting frames before or after a finite transition are not stalls"). The table above uses the corrected bounds.

**Two things the recording confirms that source could not**

- **933 ms measured against a 900 ms budget.** `TRAIL_IGNITE_SPAN_MS` is a real budget now, not a start-spread that grew with the walk. design-motion had measured the old shape at 1000 ms against a declared 1400, drifting to 1827 ms at thirty stops.
- **The settled canvas never stops** — 139 frames, minimum 0.0016, so the twinkle and the travelling light genuinely run. This is the defect that opened the whole round: byte-identical over 2.0 s with the lens open.

**Not covered.** Reduced motion and the interrupted-open case were measured in-browser (rAF sampling), not from this recording; the interruption fix was probed red/green against the reverted line and is reported in the PR body.

PR #1561, opened 2026-09-11
fix(mcp): raise the source inventory cap the dogfood repository outgrew

wlsdks:
## Summary

The `mcp` CI lane is red on `main`. `mcp/src/integration.test.mjs` connects this
checkout as a source root and asserts the handoff receipt reads
`verified_current`; it reads `review_required` with
`topGap: source_inventory_truncated`, `nextAction: review_inventory_limit`,
`missing: []`. No declared `path:` is broken — the repository crossed the bound
its own product ships.

| commit | tracked files | |
|---|---|---|
| `6da8a9c49` (#1557) | 3995 | under |
| `fb1ca40d7` (#1558) | 4007 | over the 4000 cap — this is `origin/main` |

#1558 changed no `mcp/` path, so `scripts/classify-change.mjs` never selected
the lane that would have caught it; the next PR touching `mcp/` inherited the
red.

This raises `SOURCE_INVENTORY_MAX_FILES` from 4000 to 8000 in both probes —
`mcp/src/project-source-inspection.mjs` and `src-tauri/src/lib.rs`. They are two
implementations of one measurement (the app mints the receipt, a fresh MCP
process reproduces it), so a drift makes the fingerprints differ over an
unchanged tree and reports ontology drift that never happened. There was no test
holding them equal, so `tests/contract/source-inventory-bound.contract.test.ts`
now compares all four inventory bounds plus the prune-directory set across the
two files, and also fails when this repository's own visible-file count passes
the cap — in the `unit` lane every PR selects, not only the lane a changed path
happens to reach.

`inventory-v2` is unchanged. The version string is hashed first on both sides,
and for any repository under the old cap the resulting fingerprint is
byte-identical, so no stored receipt is invalidated. The Rust truncation test's
hardcoded `"3999.txt"` now derives from the constant instead.

Per `AGENTS.md` this is a public-contract change, so `docs/DECISIONS.md` carries
one appended record: 2026-09-12, "The dogfood repository outgrew the cap its own
product ships". It cites the 2026-08-03 record ("The new MCP handoff re-verifies
the person's source connection on its own") as standing and raises one of its
bounds only. The recorded dissent is that a bounded probe which truncates and
says so loudly is the product working, and that 8000 moves the moment of
truncation rather than its shape.

## Measurements

The 2026-08-03 falsifier is "the bounded probe noticeably delays handoff on a
normal repository", so the probe was timed directly through
`inspectProjectSource`, four runs each:

| root | files hashed | cap 4000 | cap 8000 |
|---|---|---|---|
| this repository | 4000 (truncated) / 4008 (complete) | 131–142 ms, median **138 ms** | 131–144 ms, median **138 ms** |
| synthetic 8000-file git repo | 8000 | — | 114–130 ms, median **119 ms** |

The cap is a ceiling, not a workload. Cost tracks hashed bytes — already bounded
at 32 MiB by `SOURCE_INVENTORY_MAX_HASH_BYTES` — and not the file count, which
is why a full 8000-file repository of small files probes *faster* than this one
at 4008. Handoff is not delayed.

Headroom: the repository grew 3911 → 4007 files over the last 25 commits (+96),
so 8000 is roughly a thousand commits at the current rate.

## Test plan

- `pnpm integration:mcp` — **146 passed, 0 failed** (5m 10s). The named test
  `query_ontology agent_brief — read-only known-task wire path stays below 20000
  characters` was red before the change (`+ 'review_required' - 'verified_current'`
  at `integration.test.mjs:10366`) and is green after.
- `pnpm checks:changed -- --run` — **14 passed**: `source:language`,
  `docs-vault:check`, `docs:language`, `knip`, `docs:links`,
  `build && desktop:perf`, `test:desktop:bridge`, `desktop:check`,
  `test:contracts`, `test:mcp:unit`, `playwright web-surface-smoke`,
  `tsc --noEmit`, `decisions:check`, `vault:validate`.
- `cargo test --manifest-path src-tauri/Cargo.toml` — 300 passed, 0 failed;
  all five `inspect_project_source` tests pass at the new cap in 1.64s.
- `pnpm docs-vault:build` output is staged in the same commit.
- Gate probed both ways rather than trusted green: setting the Rust constant to
  9000 turns the alignment case red (`expected '9000' to be '8000'`), and
  lowering both to 3000 turns the headroom case red (`this repository has 4008
  visible files against a cap of 3000 … expected 4008 to be less than or equal
  to 3000`). Both restored.
- `cargo fmt` was run and its output inspected: the crate carries pre-existing
  formatting drift unrelated to this change (no `cargo fmt --check` gate exists),
  so only the two intended lines are in the diff and neither is reformatted by
  `cargo fmt --check`.


wlsdks:
Rebased onto `origin/main` after #1560 landed, then published as a merge rather than a rebase: `.claude/hooks/block-unsafe-git.sh` blocks `--force-with-lease` by name, so the linear history could not be pushed from here. The rebase is kept locally on `rebase-attempt-121922dc2` and `git diff rebase-attempt-121922dc2 HEAD` is empty — the tree is byte-identical, only the history shape differs.

Both #1560's and this branch's `docs/DECISIONS.md` records survive, this one above #1560's, each with its own `**Owner**` line; git had collapsed the shared one outside the conflict markers. The generated `src/entities/docs-vault/data/*` and `public/docs-vault/**` conflicts went through `pnpm docs-vault:resolve-conflicts -- --dry-run` and then its write command, per `.claude/rules/git.md` — none were hand-resolved.

Re-verified on the rebased tree: `pnpm checks:changed -- --run` 14/14, `pnpm decisions:check` clean, the new contract test 6/6, pre-push's 10 lanes ok. The repository is now 4008 tracked files against the new 8000 cap.

PR #1566, opened 2026-09-12
design(library): the pane ends with its content, touch floors reach hybrid pointers, and arrival takes input at once

wlsdks:
## Summary

Library slice S7 — three findings carried forward from the earlier slices, each measured before it was touched. Full numbers: `/Users/jinan/scratch/atlas-library-fixture-20260911/opus-s7/measurements.md`.

### 1. The source pane ends where its content ends

**Two of the brief's premises were wrong, and the measurement says so.** Across all three states at 1512×901 and 1040×720, every link from `library-source-summary` up to the shell carries `min-height: 0px`, `justify-content: normal` and `align-items: normal`; the pane's box height equals its content height (**657px** DOCX · **523px** XLSX · **777px** cited Markdown at both widths); the scroller is transparent and the ground below is `rgb(8, 9, 10)` on `main#library-page` — `--color-canvas`, the same ground the landing stands on. There is no min-height, no filler, no centring, and no second panel. The earlier reading of "395px of empty **panel**" was wrong about what the empty space is.

What was missing is the other end of the document. `pt-8` opened it and nothing closed it:

| viewport / state | last block bottom → scroll end | clearance |
|---|---|---|
| 1040×720, cited (793px of content in a 671px scroller) | 720 → 720 | **0px before** |
| 1040×720, cited | 688 → 720 | **32px after** |
| 1512×901, cited (content fits) | 772 → 901 | 129px, unchanged |

PIL on matched scroll-end frames from one build, the single declaration toggled in place: last ink **y=719 (0px gap) → y=687 (32px gap)** at 1040×720; unchanged at 1512×901. `cmp-scrollend-1040.png` shows the 「Open a copy」 door's bottom border sitting on the window's own edge before, and 32px of canvas under it after. `tests/e2e/scroll-end-gap.spec.ts` already holds a 24px floor for every other surface and `DocReadingPane` reserves its pill's clearance; this pane had the room only below `lg`.

The same row's `sha256` ran **2 × 16px lines → 1 line** at both widths (84 characters against a ~69-character mono line). The middle is elided rather than the tail, because the value is read to *compare* — against what a page recorded, or against `shasum` — and a comparison uses both ends; 16 characters per end, `truncate` behind it for widths no fixed length can allow for, and the whole 64-character value on the cell's `title`.

⚠️ Accepted side effect: at 1040×720 the un-cited DOCX pane now scrolls by 18px where it previously fit with 14px to spare. A document whose gutter does not fit scrolls; 14px was not a gutter.

### 2. The touch floor reaches hybrid pointers

`.atlas-touch-floor` / `-wide` moved from `@media (pointer: coarse)` to `@media (any-pointer: coarse)`. `pointer` reports the **primary** device, so a touchscreen laptop or a tablet with a trackpad answered *fine* and every floored control kept its 28/32px mouse height.

**Premise found wrong: there is no `any-pointer: coarse` emulation to write.** Measured in Chromium across three context shapes and four CDP calls (`setTouchEmulationEnabled`, `setEmitTouchEventsForMouse` in both configurations, `setDeviceMetricsOverride {mobile: false}`) — the moment touch exists at all, `pointer: coarse` **and** `any-pointer: coarse` both match and `any-pointer: fine` stops matching. No combination reports a fine primary pointer beside an attached coarse one. So:

- the **query** is gated in `tests/contract/touch-floor-layer.contract.test.ts`, asserting both directions because `pointer: coarse` is a substring of the new one and a loose check stays green on a revert;
- the **floor's reach** is gated in `tests/e2e/touch-target-contract.spec.ts` by a new **1280×900** touch case — a width no case in that file had ever measured. 6 floored controls scanned, 0 short of 44px. `primaryIsCoarse` is asserted `true` on purpose: it is the emulator's limit, so the day Chromium can separate the two, the assertion goes red rather than leaving a blind spot.

### 3. Arrival takes input at once

While a view transition runs, the captured document is **not painted** and therefore not hit-tested — `elementFromPoint` answers `html` everywhere and a press lands on nothing. Nobody had measured the hold.

| arriving at | committed | animations started | `finished` before | `finished` after |
|---|---|---|---|---|
| Agents | 43 ms | 56/60 ms | 301 ms | **261 ms** |
| MCP | 23 ms | 41/58 ms | 283 ms | **247 ms** |
| History | 20 ms | 39/36 ms | 278 ms | **206 ms** |
| **Library** | 22–24 ms | **229–239 ms** → never begins | **480–489 ms** | **214–238 ms** |

Press aimed at `library-open-vault`, wall-clock after the rail click: **+300 ms landed on `HTML` before, on the door after**; +700 ms and +1000 ms land in both. Fade length from first animation start to `finished`: **247 ms → 178 ms** against a declared 180 ms.

Two causes, both addressed. `-ua-view-transition-group-anim-root` and `-ua-view-transition-group-anim-app-nav-rail` each ran the UA's **250 ms**, outliving the crossfade `app/globals.css` declares, while animating a root box that a route change neither moves nor resizes and a rail whose snapshots are already excluded from the fade. And nothing bounded the transition against the arriving route's own first render, so the Library paid ~205 ms of its own work **and then** the whole animation; the hold is now bounded from the moment the route commits — one crossfade's worth of time later, a fade that has not begun never will, and it is skipped. The budget is read from `--motion-base` (which the browser reports as `0.18s`, not as authored) so the module and the stylesheet cannot drift.

⚠️ **Recorded consequence.** On the Library the fade's animation no longer runs; the screen becomes live at ~214 ms instead of ~480 ms. `pnpm decisions:find` finds no ledger record for the 2026-09-02 route crossfade — it lives in `src/shared/lib/route-view-transition.ts`'s own header, which now carries the measurement and this consequence.

⚠️ **Rejected attempt, recorded rather than deleted.** `:root::view-transition { pointer-events: none }` was built, measured, and **changed nothing** (`html` still owned every point; `finished` still 468–480 ms). The overlay is not what the hit test reaches.

## Test plan

`pnpm checks:changed -- --run <changed paths>` → **16/16 passed**, including 29 Playwright tests (the three specs below plus `overflow-sweep`), `pnpm exec tsc --noEmit`, `eslint --max-warnings 0`, `check:tokens`, `knip`, `docs:links`, `docs:language`, `source:language`, `test:contracts`, and `pnpm build && pnpm desktop:perf`. Pre-push ran 10 lanes, all green.

New and extended gates, each **probed** by planting the defect and watching it go red before restoring green:

- `tests/e2e/library-source-pane-end.spec.ts` (both widths) — removing `pb-8` → `마지막 블록과 창 끝 사이: 0px`; raising `HASH_ELISION_KEPT_PER_END` to 40 → `해시가 줄임 없이 통째로 들어갔다`.
- `tests/e2e/route-transition-input.spec.ts` — disarming the bound → `+300ms 에 누름이 문에 닿지 않았다 … {"dt":314,"on":"HTML"}`; restoring the group animations → `페이드 시작부터 끝까지 247ms — 선언값은 180ms`.
- `tests/contract/touch-floor-layer.contract.test.ts` — planting `@media (pointer: coarse)` → red.
- `src/shared/lib/route-view-transition.test.ts` — five cases on the bound (armed only at commit; skips a fade that never began; never cuts one already running; ignores animations that are not the transition's own; a second navigation disarms the first budget) plus the unit parse.
- `src/views/library/lib/elide-hash.test.ts` — equal ends, the 1040 budget, and every value it must leave alone.

**`/motion-verify` — real recording, taken.** Isolated visible Chrome launched by this session at (0,33) 1512×920 with its own profile (never the owner's browser, never the app instance another session holds), `screencapture -v -V 13 -R0,120,1512,833`, 3024×1666 VFR normalised to 30fps → 390 frames, four rail crossings. Adjacent-frame diff (share of pixels differing by >7/255):

| crossing | diff sequence |
|---|---|
| Library → Agents | 0.38 · 0.53 · **3.80** · 2.79 · 1.00 · 0.82 · 0.20 |
| Agents → Library | 0.36 · 0.72 · **2.57** · 1.83 · 0.45 · 0.30 · 0.21 |
| Library → Agents (2) | 0.09 · **3.57** · 2.64 · 2.06 · 0.84 · 0.07 |
| Agents → Library (2) | 0.36 · **2.70** · 2.19 · 0.77 · 0.40 · 0.14 |

Every crossing is a ramp across **4–6 frames (133–200 ms)** with no single-frame step and no repeated-frame freeze — a hard cut would put one frame at the peak with zeroes on both sides, which is the 2026-09-02 defect the crossfade exists to prevent. The phase strips (`phase-to-library.png`, `phase-to-agents.png`) show both screens ghosting together for two to three frames before the new one stands alone, so **the Library's arrival still reads as a crossfade** even though the bound skips its animation: the old and new snapshots are both painted while the fade waits, and the skip removes the ghost rather than replacing the screen. The `.mov` was deleted immediately after frame extraction.

**Installed-app frames deferred.** `/Applications/Ontology Atlas.app` is held by another session's instance (pid 64417, started 10:50 AM), which this slice may not drive. Browser proof on the static export stands in.

## Routing

- `pnpm po:route -- --evidence=observed --outcome=orient --change=rollback-cheap --boundary=truth:unchanged,transfer:unchanged,agent-write:unchanged,human-correction:unchanged` → `solo · door=two-way · risk=none · record=no · reviewers=none · next=build-and-verify`
- `pnpm design:route -- --change=layout,motion` → `directions=no · council=no · seats=none · proof=checks:changed, design-audit, motion-verify, computer-use-loop · sequence=build>proof`

No ledger record: `pnpm decisions:find` finds no record for the crossfade, the touch floor, or this pane, so there is none to cite or overturn.


wlsdks:
## Follow-up — the input assertion was measuring the runner, not the defect

`tests/e2e/route-transition-input.spec.ts:58` went red **3/3 retries** on the GitHub runner (`{"dt":327,"on":"HTML"}`) while passing locally. The presses were at a wall-clock offset from the rail click, and a fixed offset from the click measures how fast the machine renders the arriving route rather than how long the transition holds the screen: a slower runner commits the route later, so the bound starts later, and +300 ms lands legitimately inside a hold still waiting for a screen that has not arrived. The local before/after figures in the description stand — what changed is where the clock starts.

### Where it starts now, and why not at the door's first frame

Measured over four crossings with an in-page rAF watcher armed before the click:

| observation | value |
|---|---|
| `library-open-vault` first has a non-zero box | **+45 … +61 ms** after the click |
| the same door first answers `elementFromPoint` | **+240 … +258 ms** |
| distance between the two | **193 … 203 ms** |

Pressing at the door's first box would press **~150 ms inside the crossfade this design system chose** — it would have failed locally too, and it would assert the crossfade away rather than the hold that outlived it. The door belongs to the Library's own tree, so its first box cannot precede the route commit; that makes it a **lower bound on the commit**, observed in the page and therefore scaled to whatever machine is running, and the bound in `route-view-transition.ts` ends the hold one crossfade after the commit. So the presses are at **`doorBox + 180 ms + 4 frames` = 313 ms** and again 100 ms later. Click→layout and layout→hit-testable are logged as figures with nothing asserting on them.

| | layout → hit-testable | budget | verdict |
|---|---|---|---|
| fixed build, 9 local runs | 193–203 ms | 313 ms | **pass**, ~112 ms headroom |
| both fixes reverted (probe) | **515 ms** | 313 ms | **fail**, by 202 ms |

### The second case stopped being a stopwatch

It compared the fade's start to `finished` against the declared duration plus a frame. That measured 180–182 ms three times and then **199 ms** on the fourth — `finished` is a promise resolving on the main thread, so the number carries whatever else that thread was doing — while the defect measured **237–247 ms**. Two overlapping distributions is a coin toss, and the slack that would settle it (≥70 ms) is the slack that stops it failing on the defect.

It now reads the **animation list** captured at `ready`, before anything has started, and asserts no `::view-transition*` animation carries a duration longer than `--motion-base`. No clock, nothing to jitter, and exactly the thing that changed:

| | animations holding the transition |
|---|---|
| after | four × 180 ms on `old(root)`/`new(root)` — fade-in, fade-out, two plus-lighter blends |
| reverted (probe) | the same four **plus** `group(root)` 250 ms and `group(app-nav-rail)` 250 ms → **fail by name** |

An idling guard requires at least those four, so a run that sampled the wrong moment cannot pass on an empty set. The hold's own length is logged beside it (239–255 ms after the click on this machine).

### Verification

- the spec **3× locally green** before the merge and **3× again** on the merged tree;
- `pnpm checks:changed -- --run $(git diff --name-only origin/main...HEAD)` → **14/14**;
- pre-push 10 lanes green;
- two lint warnings the rewrite introduced (an unused `layout` read and the retired `SLACK_MS`) were caught by `lint-warning-ratchet.contract.test.ts` at 0 → 2 and **removed**, not absorbed into the ceiling;
- branch merged with `git merge origin/main` and with this PR's own updated tip — no force-push.

PR #1584, opened 2026-09-12
fix(library): the picture fills its canvas, the fit tile moves, and every mark keeps its name

wlsdks:
## Summary

Four findings from inspection 122 on the installed `23cbf6d64` build, all on the Library home graph. Three of them are the same kind of defect: a mark or a control that says something the canvas does not.

**S1 — the picture used a third of its canvas and 「맞춤」 could not fix it.** The camera was clamped to `[0.4286, 1.6]`, and the small folders sat *against* that ceiling at every window — so the fit had nothing left to take and the tile's press was a no-op (measured: 0 changed pixels). The ceiling is now stated as **the widest mark a folder may draw — 36px, the map's own node chrome, never wider** — and `libraryZoomMax` turns it into a scale: 2.0 for an ordinary folder (was 1.6, a 28.8px page), up to 3.6 for one whose marks all sit at the band's floor. The fit tile reads `isSameView` against the fit the camera would take and takes `ChromeTile`'s disabled grammar while the picture is already framed, so no press can answer with the same pixels.

**The brief's 60–75% of the shorter axis is deliberately not taken.** On this folder that needs a camera of ≈2.4 and a **43px** page — the balloon `docs/DECISIONS.md` 2026-09-12 ("One folder is one picture") records the owner rejecting — and being window-relative it restores that record's own falsifier, *a folder whose picture changes shape or size with the window*. The fill therefore stops where the mark cap stops.

| folder | window | camera | fill W % | fill H % | tallest empty band | widest mark |
|---|---|---|---|---|---|---|
| `vault` | 1512×901 | 1.60 → **2.00** | 38.4 → **45.9** | 38.7 → **47.8** | 257 → **222** | 28.8 → **36.0** |
| `vault` | 1040×720 | 1.60 → **2.00** | 67.8 → **81.0** | 52.4 → **64.7** | 153 → **117** | 28.8 → **36.0** |
| `vault` | 1920×1080 | 1.60 → **2.00** | 27.9 → **33.4** | 31.0 → **38.2** | 357 → **321** | 28.8 → **36.0** |
| `vault-zero-answers` | 1512×901 | 1.60 → **2.00** | 39.5 → **47.3** | 32.8 → **40.4** | 281 → **251** | 28.8 → **36.0** |
| `vault-zero-answers` | 1040×720 | 1.60 → **2.00** | 69.7 → **83.5** | 44.5 → **54.7** | 176 → **147** | 28.8 → **36.0** |
| `vault-zero-answers` | 1920×1080 | 1.60 → **2.00** | 28.7 → **34.4** | 26.3 → **32.4** | 380 → **351** | 28.8 → **36.0** |
| `vault-60` | 1512×901 | 1.25 → 1.25 | 53.9 | 88.4 | 59 | 22.5 |
| `vault-60` | 1040×720 | 0.867 → 0.867 | 79.2 → **86.1** | 83.6 | 60 | 15.6 |
| `vault-60` | 1920×1080 | 1.60 → **1.62** | 52.4 → 53.0 | 91.9 → 93.0 | 45 → **40** | 28.8 → 29.2 |
| `vault-300` | all three | unchanged | unchanged | unchanged | unchanged | unchanged |

The two bigger folders fit below the old ceiling already, so only their names change; grid occupancy on the two small ones moves 0.29 → **0.42** and 0.42 → **0.71**.

**S2 — the amber "unverified citation" dot.** The cause was neither zoom nor dot size: the dot was drawn **only while `flow.pulse` held the edge** — one breath of two settle budgets as the home arrives, and never again, or while a card stands open on one of its ends. At rest, which is where the home lives, `flow` is `null` and nothing amber was painted at all; the inspection's two captures fell on either side of that single breath. And even mid-breath its resting 0.55 alpha composites to `rgb(139,105,33)`, which is not this product's warning amber. So the fix is both: every unverified citation keeps the dot **at rest**, at the line's own ink, inside half the break it sits in and floored at 3px across, and the pulse adds radius rather than brightness. Measured with PIL over the saved canvases (`R>190, 140<G<210, B<120`): `vault` at 1512 **0 → 99**, at 1040 **0 → 91**.

**S18 — one mark with no identity.** A name that loses all four of its places is now pushed onto 12 spokes × 4 steps out to 44px, slid inside the frame, tied back by a hairline leader in its own ink — and only a name that has already lost its four places pays for the search, which is skipped above 120 marks. Marks carrying a name and drawn without one: `vault` and `vault-zero-answers` **1 → 0** at all three windows; `vault-60` **6 → 1 / 2 → 0 / 1 → 0**.

**S19 — the stale legend.** Two stale sentences, not one: with a mark active the line was that mark's description ending in 「누르면 옆에 카드가 열립니다」, and with nothing active the legend ending in 「점을 누르면 그 옆에 카드가 열립니다」. Both keep every fact about the marks and the lines and swap only their **gesture clause** for 「Escape 또는 다시 누르면 닫힙니다」 (`legendCardOpen`, `describeOpen.*`). The card's amber sentence is printed from that mark's own unverified citations, which are now exactly the edges carrying a drawn dot.

`docs/DECISIONS.md` gains one record narrowing 2026-09-12 "One folder is one picture" on its own design-lead dissent, with the 43px arithmetic as the standing refusal and design-infoviz's objection (120 standing dots at 372 marks) kept.

## Test plan

- `pnpm checks:changed -- --run <16 changed paths>` — **16/16 passed**, including `pnpm build && pnpm desktop:perf`, `pnpm test:contracts`, `pnpm exec tsc --noEmit`, `pnpm test:i18n:messages`, `pnpm decisions:check`.
- `pnpm exec playwright test tests/e2e/library-graph-picture.spec.ts` — **16 passed**, four folders × three windows plus the two new interaction cases.
- Pre-push lanes: all 10 green (`contract` 275s, `unit` 110s).
- The legibility gate gains five bars, **each planted red before it was trusted**: the fit takes the camera it is allowed (`Expected: 2 · Received: 1.6` with the old ceiling), an amber pixel exists at rest (`Expected: > 0 · Received: 0` at all three windows with the pulse-only dot), no anonymous mark under 30 marks (names `source:sources/settlement-policy.md` — the exact mark S18 photographed), at most one under 120, the fit tile only offers a press that moves the picture (`Expected: disabled · Received: enabled`), and the legend answers the open card.
- Frames, unchanged fixtures: 372 marks with a card open 0.69 → 0.83ms, 992 marks 1.52 → 1.86ms — both under the 2ms budget the card record staked.
- **Rendered evidence is from the browser, not `/Applications/Ontology Atlas.app`** — the brief for this change forbids driving the installed app, so nothing here was captured from it. Canvas and window PNGs plus per-case JSON for all 12 cases, before and after, and the full write-up: `/Users/jinan/scratch/atlas-library-fixture-20260911/opus-f3/measurements.md`.
- `pnpm design:route -- --change=layout` returned `directions=no · council=no · proof=checks:changed:changed-paths, design-audit:affected-state, computer-use-loop:affected-state`; the loop was run in Chromium at 1512×901, 1040×720 and 1920×1080.


wlsdks:
---

## Follow-up commit: the third lever, measured then refused

The camera is against its ceiling, so the remaining way to make a small folder's picture bigger is to grow the **world** — the picture grows, the marks do not. Built and measured over the same 4 fixtures × 3 windows; refused on its own numbers, with the refusal turned into a gate bar.

**Not the springs.** `CITES_REST` was cut 52 → 28 in G2 because at 52 *"a page and the six files it was written from occupied as much of the picture as three unrelated pages did: the hub and its satellites read as a constellation of equals rather than as one thing"* — measured on a folder of exactly the size S1 is about. A rest length is a claim about meaning and may not be spent on fill.

**So the candidate was the gap between unrelated groups** (`ORPHAN_RING_GAP`, the packing gutter), scaled by mark count alone and never by the window, so `vault-60` and `vault-300` measure byte-identical.

| gap × | `vault` fill at 1512 | 6×4 occupancy at 1512 | tallest band | group spread | camera / widest mark at **1040** |
|---|---|---|---|---|---|
| 1.00 (shipped) | 45.9% × 47.8% | **0.417** | 222 | 0.698 | 2.00 / **36.0px** |
| 1.15 | 47.8% × 49.8% | 0.375 | 214 | 0.712 | 2.00 / 36.0px |
| 1.28 (the cap) | 49.5% × 51.5% | 0.375 | 207 | 0.720 | 2.00 / 36.0px |
| 1.55 | **59.8%** × 51.4% | 0.375 | 207 | 0.751 | **1.74 / 31.4px** |

1. **60% arrives only at 1.55, and costs one folder two sizes.** There the 1040×720 canvas (616×590, inner 488×462 after the fit's padding) can no longer hold the picture at the ceiling: the camera drops to 1.74 and the same twelve marks are **31.4px on one window and 36px on another** — the G1 observation verbatim (*"26.1px at 1040 and 34.0 at 1920"*). The arithmetic is not close: 60% of the 1512 canvas is 653px of picture and the 1040 canvas has 488px to give.
2. **Below that, every step buys bounding box while losing occupancy.** The 6×4 grid the packing record improved (0.54 → 0.83) reads 0.417 today and 0.375 at every stretch — the extra area is air between groups, not picture. The statistic in the complaint improves while the better one gets worse: the fill objective G2 deleted, re-entering through a gap.

Kept instead: **a folder of 30 marks or fewer draws its widest mark at exactly `LIBRARY_MAX_MARK_PX` at every window** — the bar that makes this refusal enforceable and is the whole of what an absolute ceiling was for. Probed red with the 1.55 build: 1512 green at 36.0px, 1040 red at `Received: 31.36648174841914`. The numbers and the reasoning also sit in `library-force-simulation.ts` beside the gutter, so the next pass does not re-run the experiment.

PR #1711, opened 2026-09-19
fix(acp): a call whose turn ended stops claiming it is running

wlsdks:
## Summary

Measured in the rendered dock. Press **Stop** while a read is in flight, or let
a turn end without the adapter closing its call, and the transcript row kept the
word `running` and its pulsing indigo ring **for the rest of the conversation** —
with the session back at `Ready` and the composer open.

| moment | row | ring |
|---|---|---|
| while the read is in flight | `read_source · sources/architecture.docx` · **running** | yes |
| after Stop (before) | `read_source · sources/architecture.docx` · **running** | **yes** |
| after the turn simply ended (before) | `read_source · sources/architecture.docx` · **running** | **yes** |
| after Stop (after) | `read_source · sources/architecture.docx` · **never finished** | no |

Nothing on the wire says otherwise, which is why the row could not tell: a
cancelled or crashed adapter simply stops reporting, so the last thing the call
ever said about itself was `pending`. Only the panel knows whether that call is
still open or was abandoned there.

## What changed

`readToolOutcome` takes the turn's state as a **required** argument — not
defaulted, for the same reason `fromVaultServer` is not: either default is a lie
in one direction, and a forgetful caller would print the lie.

- An open call whose turn is over reads `unfinished` / `끝나지 않음`. That is the
  **absence of an ending, not a failure** — nobody said it failed, so it gets
  neither the danger seam nor a count, and the word carries it in the
  right-aligned column the way `nothing found` already does.
- The ring follows the word it already meant. Its own comment says *"The dot
  means 'still going', so a finished row has none"*; a stopped call now has none.
- A call is live only while the session is thinking **and** the call opened in
  the current turn. Membership of the current turn is the half that cannot be
  skipped: a row left open by an earlier turn would otherwise start claiming
  `running` again the moment the next question was asked.

### It makes an existing promise true

`stoppedWithoutAnswer` (landed earlier today as #1664) deliberately stays quiet
when a tool ran, because *"a tool that was mid-flight carries its own stopped on
its row"*. The row was carrying the opposite word, so a stopped turn that had
reached a tool said nothing true anywhere. It does now.

## Test plan

`pnpm checks:changed -- --run` — 9 checks, all green: eslint, both sibling Vitest
files (146 cases), `source:language`, `knip`, `test:contracts`, `tsc --noEmit`,
`test:desktop:check`, `test:i18n:messages`.

New unit cases in `tool-outcome.test.ts`: an open call reads `unfinished` once
its turn is over; a **finished** call is never turned into an unfinished one by
the turn state; and an unfinished call refuses a count the same way a broken one
does. Two existing cases that emitted a `tool_call` with no live turn now start a
real turn first — a tool call only ever arrives inside one.

New panel case: Stop pressed on an in-flight read leaves a row with no ring, a
`data-tool-outcome` of `unfinished`, and no `running` anywhere in it.

Rendered proof: the Library dock driven through `installLibraryWorkHarness`,
both the Stop path and the completed-turn path, before and after.


wlsdks:
This landing has held the repository's landing lock for 36 minutes without merging, and from the outside everything it waits for is already done.

Observed just now:
- All 16 check runs on the head commit `24122bf` are green, including every one of the 8 required contexts.
- `mergeStateStatus` is CLEAN and the pull request is mergeable.
- The lander process is alive (pid 32213), so the lock is refreshed on every poll and the 45-minute lease never expires. A live-but-stuck lander holds the repository indefinitely, which is different from the crash case the lease was designed for.

Nothing else can merge while this holds. If the process is stuck rather than working, stopping it releases the lock and the landing can be restarted cleanly; if it is genuinely mid-merge, ignore this.

One thing that may be a red herring but is worth ruling out: the legacy combined-status endpoint for that commit returns `pending` with zero statuses, because this repository reports through check runs only. If any wait path reads the combined status rather than check runs, it would wait forever on a commit whose checks are all green.

wlsdks:
Correction to my comment above: the combined-status theory is wrong, and I should not have offered it.

`pr-land.mjs` never reads the legacy status endpoint. It judges required contexts off `statusCheckRollup` from `gh pr view` (PR_FIELDS), and there is no combined-status call anywhere in the file. I checked after main-7-c2 pointed it out. So a commit whose checks are green but whose combined status reads "pending with zero statuses" would not hang this wait path, and that observation explains nothing.

What still stands is the part that does not depend on a theory: all 16 check runs on 24122bf are green, all 8 required contexts included, the pull request is CLEAN and mergeable, and the lock has been held without merging for about 40 minutes. The lease cannot end it, because the holder refreshes the lock on every poll.

The line that would actually decide it is in this lander's own log — "waiting on required checks: N running, M not reported yet" — which separates contexts still running from contexts that never reported. I could not read it from here: its stdout runs through a pipe chain and the captured file is empty. I have told the session that owns the process.

PR #1715, opened 2026-09-19
fix(acp): a fenced block that runs past the panel says so

wlsdks:
## Summary

Measured in the rendered dock. A command in an answer drew as

```
pnpm atlas compile --page wiki/archite
```

at a **320px panel** — cut flush at the right edge, with no fade, no rule and no
resting scrollbar, because a WebView's overlay scrollbars appear only while the
pointer is moving.

A truncated sentence is visibly truncated. A truncated **command looks
complete**, and copying it runs the wrong thing.

The block now wears the affordance this repository already uses for exactly this
fact: the `--tabbar-edge-fade` mask the tab strips, the transcript's own top edge
and the past-conversation list all carry. No new token, no new width, the same
four states. The markdown table's scroller gets it too — same reason, same place.

**Only while something is actually hidden.** Painted unconditionally the mask
would blur the end of a command that fits, which states the opposite of the fact
it exists to state — the rule the transcript's top fade already keeps.

| panel | hidden | edge | mask |
|---|---|---|---|
| 520 (the default) | 0px | none | none |
| 320 (the documented floor) | 64px | `end` | right edge fades |
| 320, scrolled to the end | 64px | `start` | left edge fades |

## The gate

`tests/e2e/chat-answer-overflow-edge.spec.ts`, four cases.

⚠️ **Both states are measured at a width where that state exists.** A gate that
only ever looked at the narrow case would never catch a mask painted
unconditionally, so the same answer is read at the width where the command fits
and at the width where it does not — and the command is the one measured in the
dock, unchanged. A longer one would prove only that a very long command
overflows everywhere.

**Deliberate RED, precisely.** Neutering only the edge measurement (`start`/`end`
forced false) fails exactly the two cases that assert something *is* hidden,
while the two that assert nothing is hidden keep passing. Restoring returns 4/4.

## Test plan

`pnpm checks:changed -- --run` — 7 checks, all green: eslint, the new Playwright
spec, `AcpChatPanel.test.tsx`, `source:language`, `knip`, `test:contracts`,
`tsc --noEmit`.

Rendered proof: a Korean answer in the Library dock at 520 and 320 through
`installLibraryWorkHarness`.


wlsdks:

PR #1719, opened 2026-09-19
fix(acp): the usage-limit card stops naming a picker that is not there

wlsdks:
## Summary

The error card must not name an action it does not offer. That rule is why Retry
lives inside this card at all (owner, installed app, 2026-08-24: *"if this is
normal I still would not know what to do"*). The usage-limit sentence had quietly
broken it again:

> Not the app: the tool's own plan is out of room. If Details names when it resets, press Retry then; **until then pick another tool below and carry on.**

Measured with one tool installed — the ordinary case, and the one that actually
hits a usage limit:

| on screen | |
|---|---|
| runtime picker | **none** — the footer shows the tool's *name* |
| mode picker | none |
| composer | disabled |
| controls | Retry · connection check · New chat · Send |

There was nothing to pick, and nothing said so.

## The fix

A second tool is exactly what `runtimes` / `onRuntimeChange` already decide the
footer picker by, so the sentence is now chosen by **the same fact the control
is**. Nothing is added to the screen; one of the two sentences is simply true.

- **One tool** → says so, and names where a second one is connected — the Agents
  screen, a real destination rather than an absent picker.
- **Two or more** → the original sentence, which was right all along, and the
  picker it points at is there.

## The gate

`tests/e2e/agent-limit-card-next-step.spec.ts`, four cases.

⚠️ **Both states are measured**, because the sentence is only wrong in one of
them and a gate that looked at the other would stay green forever. One case also
asserts the negative directly: with one tool, no control on the panel can switch
tools.

**Deliberate RED, precisely:** removing only the sentence selection fails exactly
the one-tool case; the two-tool case keeps passing, because that sentence was
never wrong. Restoring returns 4/4.

## Test plan

`pnpm checks:changed -- --run` — 9 checks, all green: eslint, the new Playwright
spec, `AcpChatPanel.test.tsx`, `source:language`, `knip`, `test:contracts`
(2729), `tsc --noEmit`, `test:desktop:check`, `test:i18n:messages`.

Rendered proof: the Library dock with a rejected `acp_start`, before and after.


wlsdks:

PR #1729, opened 2026-09-19
design: the product speaks one Korean

wlsdks:
## Summary

Owner, 2026-09-19, of the Korean catalogue: some of it reads like a bad
translation. Measured, it was speaking in **two registers at once** — 839 of
5,055 strings ended formally and the rest plainly — and the split ran **through**
single surfaces rather than between them.

- The permission card's own title was plain while the review inside it was formal.
- One notice managed both **inside a single sentence**:
  `{detail}을(를) 묻지 않고 **썼습니다**. 위키 서식에 **맞아서요**.`
- The Library alone carried 159 formal strings beside its plain ones.

All 839 are now the plain register — what every first-contact surface already
used: the empty states, the status words, the first-run wait, the buttons.

## It is not a blind substitution

Korean conjugation contracts, so a regular rule produces wrong words. An explicit
hand-written table carries **all 97 irregular endings the catalogue actually
contains** — the everyday verbs for *becomes*, *is not*, *writes*, *makes*,
*differs* — and the regular rule handles only what is genuinely regular, with
vowel harmony. Anything neither could reach was reported rather than guessed;
that set came to five.

Those five ended in a copula **after a runtime value**, where the correct ending
depends on whether the value ends in a consonant and nothing can know that in
advance. They are rewritten so they no longer need one.

### I under-reported this earlier

My first count said 605. It was wrong: it looked only for the three most common
spellings and could not see 97 forms that are ordinary verbs. The real number is
839 strings / 1,120 endings. The gate matches the formal terminator broadly for
exactly that reason — a narrow matcher does not make the catalogue cleaner, it
makes the gate blind where nobody thinks to look.

## Three gates were pinning prose, and went red on correct Korean

`documentation.md` already names this failure: *check a fact a machine can
derive, never a sentence a person wrote.* Two now match the fact — a build is
not code-signed, the app has no account and uploads nothing — and two literal
expectations were updated.

## The borrowed-word rule stays scoped to the panel, and the scoping is a finding

Written catalogue-wide it reddened on 24 strings that are **not** the same
defect: the Library counts work in `에이전트 턴`, a named unit it prints fifteen
times and tallies (`에이전트 턴 1회`, `에이전트 턴 없음`); the native Git errors
call a commit a `스냅샷`; two download captions describe a screenshot as `캡처`.
Those are deliberate vocabulary on surfaces this change did not measure, and
widening a panel-local judgement over them would be this gate overruling a
product decision it never looked at.

## Test plan

`pnpm checks:changed -- --run` — 8 checks green: `check-desktop-readiness`,
`validate-messages`, `source:language`, `knip`, `test:desktop:check`,
`desktop:check`, `test:contracts`, `tsc --noEmit`.

Pre-push unit lane: **912 passed**. 25 component assertions pinned the old
wording; each is repinned to the same sentence in its new register, so a real
copy regression still reddens them.

**Deliberate RED:** putting one sentence back into the formal register fails the
gate and names the key.

## Noted, not done here

`src/shared/lib/validate-vault-document.ts` and a few siblings emit Korean
sentences **directly from source** rather than through the catalogue, so they are
outside both this change and the gate. That is its own finding.


wlsdks:
Closing in favour of a fresh pass against today's main.

This branch unified the Korean register as of 2026-09-19. Since then 15 commits have touched `messages/ko.json` (#1704, #1787, #1788, #1789, #1721 among them), so the branch is 40 commits behind and conflicts in 28 places. Two things made re-doing the cheaper and safer option:

- Of the 26 conflicting strings in `ko.json`, some would revert copy that landed afterwards — `"비즈니스 흐름"` against main's `"제품 흐름"` from #1704, for one. Resolving them one by one means re-deciding each against newer copy that was already reviewed.
- Even resolved, the unification would not cover the strings added in those 15 commits, so the branch cannot deliver "one Korean voice" any more.

The direction this branch chose is kept: **해요체**. Measured on today's main, it is already the majority — 597 해요체 against 474 합쇼체 among strings that end in a verb — so the new pass converts the minority rather than the majority, and this branch's own resolutions agreed with that. Nothing is dropped; the work is being redone at a scope that can actually be complete.

Branch tip recorded: `514b29b6c`.

PR #1733, opened 2026-09-19
design(map): what the ego focus dims recedes to one rest alpha

wlsdks:
Stacked on #1728 (which stacks on #1725); their commits are included until they land.

## Summary
- **The dim state was invisible.** Ego focus dims non-neighbours by colour only: measured `--map-node-stroke-dim` #6a6a74 against the capability stroke #80808c is **1.37:1**, and the dim fill #1e1e24 is *lighter* than the normal fill #17171d. Selecting 교환 접수 drew its three sibling capabilities as bright, nameless circles indistinguishable from the ego members.
- **One number for what the focus dims.** Dimmed nodes now sink to `--map-ego-rest-alpha` (0.42) on the node's own focus ramp (`egoRestSink`), applied in the effective-alpha pass so lines, chips and the `nodes()` probe follow. A dimmed project or domain keeps its name at that same alpha (its node's), a dimmed child none. `--map-ego-dim-label-alpha` is retired: it was the same 0.42 on a second name.
- Lenses (recent/path/constellation/trail) keep their own sink, untouched.
- Docs: DESIGN-SYSTEM text ladder, FEATURES focus bullet; the FEATURES line claiming a double-click "local graph" mode (2-hop, breadcrumb) is corrected to what double-click does since #1720.

## Test plan
- [x] RED → GREEN e2e `tests/e2e/map-ego-rest-alpha.spec.ts`: token present; three siblings and another domain at 0.42; focus, neighbour and parent stay well above. RED on the untouched tree (token absent, siblings at 1).
- [x] Unit: `focus-state.test.ts` (`egoRestSink` ramp/clamp), `labels.test.ts` (dimmed domain follows its node's alpha; dimmed child 0), `read-map-tokens.test.ts` (token rename).
- [x] `pnpm checks:changed -- --run`: 16/16 (includes `map-ego-dim-labels`, `overflow-sweep`, `check:tokens`, `design:toc:check`).
- [x] Siblings: path lens sink/frame, spotlight clearance, edge hover card, expand-all, 3D return, galaxy transition, cross-domain neighbours, double-click, smoke — 17 passed, 1 known dev-server flake (doc back-navigation) passed on retry.
- Browser after: focus/neighbour/parent 1, siblings and other domains 0.42, camera gap 0.


wlsdks:

PR #1734, opened 2026-09-19
design(map): INDEX marks where the query landed

wlsdks:
Independent of the search-palette stack (#1726, #1727) — different files, branches from `main`.

## Summary

The map's INDEX panel has its own search field. It filtered 125 concepts down to two and **marked nothing in either**, on an 11px label — while the `⌘K` palette on the same screen had been marking its rows for months. The filter said which rows survived; nothing said where the word was.

Both index panels (`TopologyIndexPanel`, `TopologyRealmLedger`) now hand their trimmed query to `TopologyIndexTreeRow`, which draws the name through `HighlightedText`. A chosung query marks there too, because `shared/lib/highlight-match` already reads a Hangul keyboard (#1724).

### Measured on the map panel's surface, not the palette's

The mark is tokenised for the palette's surface, so it was re-measured where it now renders: ink `rgba(210,218,255,.98)` over `rgba(139,151,255,.22)` composited on the panel's `rgb(23,23,28)` → **8.59:1**.

| typed into INDEX | marks before | marks after |
|---|---|---|
| `주문` | 0 | 11 |
| `정책` | 0 | 3 |

## Left open, with its measurement

Rows kept only as the **path** to a match stay unmarked — which is what they are. But so do rows matched through the id slug when the name is Korean: `order` leaves 20 of 20 rows unmarked, `cart` 4 of 4. In the palette that was answered with a reason column (#1726); a tree row at this density has no seat for one, so telling a match apart from the path it sits on is its own design question rather than a transplant.

## Test plan
- `TopologyIndexPanel.test.tsx` — a filtered row carries a `<mark>` with exactly the typed text; an empty query renders none. **Gate-probed**: removing the `HighlightedText` turns the first red and leaves the other 45 green.
- Ran: `eslint --max-warnings 0` on the four changed files · `vitest run src/widgets/topology-index-panel` (7 files, 82 tests) · `pnpm source:language` · `pnpm knip` · `pnpm tsc --noEmit` · `pnpm test:contracts` (272 files, 2729 tests) · `pnpm docs:check`.
- Live on `/ko/topology` at 1440×900 against the running build, with the contrast measured in the page.


wlsdks:

PR #1735, opened 2026-09-19
fix(map): a name under the floor band goes above its node instead of vanishing

wlsdks:
Stacked on #1733 → #1728 → #1725; their commits are included until they land.

## Summary
- **Two domains at the canvas floor drew nameless.** The label anchor gate drops any label whose slot below the node falls under the bottom safe band (72 px at ≤880 tall, 96 otherwise). That band exists for the readout in the bottom-right corner (measured DOM: x 1171–1488 only), yet the gate applied across the whole width. In ego focus at 1512×806, 마케팅 and 적립·멤버십 sat at y 741 of 806, inside the picture, and lost their names while every other domain kept its.
- **The name takes the slot above.** `floorFlipBaseline` (in `render/label-layout.ts`) returns the flipped baseline when the slot below is under the band, the node is on screen, and the slot above is inside the safe rect; the frame uses it as the candidate's baseline. The ordinary gate and the protected-label clamp are unchanged for every other case.
- `docs/FEATURES.md` focus bullet and a change record.

## Test plan
- [x] RED → GREEN e2e `tests/e2e/map-floor-labels.spec.ts` at 1512×806: every visible domain named; the low ones named above their node, inside the canvas. RED on the previous tree (2 nameless domains).
- [x] Unit: `label-layout.test.ts` `floorFlipBaseline` (flip; in-band untouched; node itself off-canvas; slot above unsafe).
- [x] Siblings: spotlight clearance, ego dim labels, ego rest alpha, cross-domain neighbours, edge hover card, expand-all, short viewport, 3D return, galaxy, path lens frame — 16 passed.


wlsdks:

PR #1736, opened 2026-09-19
fix(search): the palette says how many it found, not how many fitted

wlsdks:
**Stacked on #1727 → #1726 → #1724 (landed).** GitHub retargets each as the one below lands.

## Summary

Each group draws at most 20 rows, and **both the group heading and the footer counted the drawn array** — so the limit stood in for the answer. The footer is the one place that names the scope it searched ("N matches · Online Store"), which makes its number read as a fact about that folder rather than about the page.

Measured on the bundled Online Store sample (125 concepts):

| typed | heading before | footer before | actually matched |
|---|---|---|---|
| `the` | 매치 · 20 | **21 매치** | **120** |
| one common Hangul initial | 매치 · 20 | 20 매치 | **52** |
| `a` | 매치 · 20 | 21 매치 | **126** |
| `주문` | 매치 · 11 | 11 매치 | 11 ✓ |

The empty state already had this right — it has shown "20 / 125" for months. The query state simply never used the shape the product already knew.

### After

| typed | heading | footer |
|---|---|---|
| `the` | 개념 · 매치 · **20 / 119** | **120 매치** |
| one common Hangul initial | 개념 · 매치 · **20 / 52** | **52 매치** |
| `주문` | 개념 · 매치 · 11 | 11 매치 |

The matchers return the page **and** the size of what was found (`{ results, total }`). The heading shows "shown / found" wherever the cut happens; the footer carries the real total. Below the limit nothing changes.

## Test plan
- `match.test.ts` — the limit test now asserts `total` as well as `results.length`; an empty query counts the filtered whole (and with a kind filter, that filter's whole); under the limit `total === results.length`; projects report their total the same way.
- The existing ~40 assertions are unchanged — the call sites destructure `{ results: r }`, so what each case asserts stayed put.
- Ran: `eslint --max-warnings 0` on the three changed files · `vitest run src/widgets/global-search` (76 tests) · `pnpm source:language` · `pnpm knip` · `pnpm tsc --noEmit` · `pnpm test:contracts` (272 files, 2729 tests) · `pnpm docs:check`.
- Live on `/ko/topology` at 1440×900 against the running build; the before/after tables above are that measurement.


wlsdks:

PR #1738, opened 2026-09-19
fix(search): Enter belongs to whichever control has focus

wlsdks:
**Stacked on #1736 → #1727 → #1726 → #1724 (landed).**

## Summary

cmdk's root listens for Enter across the whole palette and turns it into "open the highlighted row", `preventDefault` included — so it also swallowed Enter pressed on a **control**.

Measured live on `/ko/topology`:

| focus | pressed | what happened |
|---|---|---|
| a kind filter chip | Enter | chip stayed `aria-pressed="false"`; the palette **closed** and the map **flew to `capability:account-closure`** — whichever row happened to be highlighted |
| the close button | Enter | **navigated** to the highlighted concept instead of closing |
| the same chip | Space | nothing stolen — Space would have toggled it |

So the two activation keys disagreed about what the focused control does, and the one people reach for first did something destructive to their context.

Enter now stops at the control's own row in the **bubble** phase, so the button still receives it and cmdk's root is left with the Enter that comes from the search field — the one place "open the highlighted row" is what a person means.

### After, with real key events

| focus | pressed | result |
|---|---|---|
| kind chip | Enter | `역량=true`, palette open, list narrows to `20 / 54`, no navigation |
| close button | Enter | closes, no navigation |
| search field | Enter | opens the highlighted row, as before |

A **synthetic** `KeyboardEvent` never triggers a button's native activation, so it can only prove nothing is stolen — the positive half was verified by dispatching real key events through the browser.

## Test plan
- `GlobalSearch.test.tsx` — three cases: Enter on a filter chip does not call `onSelectNode`; Enter on the close button does not either; Enter from the field still calls it exactly once. **Gate-probed**: removing the guard turns the first two red and leaves the third green.
- Ran: `eslint --max-warnings 0` on both changed files · `vitest run src/widgets/global-search` (79 tests) · `pnpm source:language` · `pnpm knip` · `pnpm tsc --noEmit` · `pnpm test:contracts` (272 files, 2729 tests) · `pnpm docs:check`.


wlsdks:

PR #1740, opened 2026-09-19
fix(acp): nothing on the footer is drawn over anything while a turn runs

wlsdks:
## Summary

Measured on the rendered footer with a live turn. A running turn puts the **Stop**
chip on the right-hand group, which takes it to **215 of the 228 pixels** the
footer has at the panel's documented minimum. The picker row was left with
**nine**, and the picker inside it does not shrink below its floor — so it drew
**on top of the status word**.

```
status "Thinking · 1s"  hits: [acp-chat-mode, svg, acp-chat-mode, self, self]
```

Three of five probe points over the status came back as the picker or its
chevron.

⚠️ **The footer's `scrollWidth` equalled its `clientWidth` the whole time.**
Nothing measuring rectangles would have seen this — occlusion is not geometry,
so the gate probes every visible piece with `elementFromPoint` across its own box.

From 320 to 380 the same squeeze also **cut the picker's word**, which is the
promise the runtime name already yields to keep (#1705).

| panel | footer | mode, before | status covered, before | after |
|---|---|---|---|---|
| 320 | 228 | 48px over a 74px word | **3 of 5 points** | stands down · 0 |
| 360 | 268 | 49px | 0 | stands down · 0 |
| 380 | 288 | 67px (cut) | 0 | stands down · 0 |
| 400 | 308 | 74px whole | 0 | **74px whole** · 0 |
| 436 | 344 | 74px whole | 0 | 74px whole · 0 |

## What yields, and why that one

The mode stands down **while a turn runs, and only then**. During a turn it is the
one control on the row that cannot act — the mode was fixed when the turn started
— while the status and its clock are live information and Stop is the way out. It
comes back the moment the turn ends, which the gate checks rather than assumes.

308 is the footer width at which the Korean mode name and the running group both
measured whole. Korean decides, as it did for the name.

`pnpm design:route -- --change=responsive` → `directions=no · council=no`.

## Test plan

`pnpm checks:changed -- --run` — 7 checks green: eslint, the new Playwright spec
(8 cases), `AcpChatPanel.test.tsx`, `source:language`, `knip`, `test:contracts`,
`tsc --noEmit`.

The gate covers both states at the widths where each exists, in both locales:
nothing covers anything while busy at 320 and 380; the mode keeps its **whole**
word while busy at 400; and at the minimum width it is present when idle, hidden
while running, and back — whole — when the turn ends.

**Deliberate RED:** removing only the condition fails 6 of 8, naming what is drawn
over what.


wlsdks:

PR #1741, opened 2026-09-19
design(map): the expand bar names its own scope, not the toolbar's

wlsdks:
Branches from `main`, independent of the #1725 → #1728 → #1733 → #1735 stack.

## Summary
- **Two controls, one name.** The map toolbar's button (`searchWidgets.hint.expandAllLabel`) opens every folded group and refits; the bar above a selected parent (`topology.cluster.barExpandAll`) opens that parent's children. In English both were the literal string `"Expand all"`; in Korean, two synonyms of it. Measured on the sample map: selecting the payments domain put one of them over the node while the other sat in the toolbar, with nothing saying which stays inside the node.
- **The bar takes the bare verb** (`topology.cluster.barExpand` — "Expand", and its Korean equivalent), the mirror of its own Collapse. The counted form `Expand {count}` is unchanged, and so is the rule behind it: a number appears only when the batch is smaller than the fold, because the node engraves its total. That rule and the verb-button shape come from the 2026-08-02 ledger decision, which this narrows rather than overturns — decision fragment included, with the dissent (the bare verb drops the "everything opens" promise) kept as the falsifier.
- New gate: a contract reads both locales and fails if **any** `topology.cluster` string equals the toolbar's expand-all label.

## Test plan
- [x] RED → GREEN contract (`tests/contract/expand-settings.contract.test.ts`): 4 failures on the original source — the bar's label, the no-repeated-number rule, the reservation width, and the new two-names check — all green after. The count-only-when-batched rule is re-asserted, not weakened.
- [x] `pnpm checks:changed -- --run`: 15/15, including the full contract suite (2,731 tests), `test:i18n:messages` and `docs:language`.
- [x] Siblings: `map-expand-all`, `map-smoke`, `map-edge-hover-card`, `map-spotlight-label-clearance` — 9 passed.
- [x] Browser after: the bar over the payments domain reads the bare verb above a node engraved `15`; the toolbar keeps its own words.

## Also in this round (no change needed)
Hover, keyboard walk, expand-all at 125 nodes (0 label or node overlaps), the INDEX tree bars (each fill is its total over the largest, ±0.2 px) and the recent-changes dialog were measured and left alone.


wlsdks:

PR #1742, opened 2026-09-19
fix(search): the reason survives a phone

wlsdks:
**Stacked on #1738 → #1736 → #1727 → #1726 → #1724 (landed).**

## Summary

The reason column added in #1726 was `md:`-only, so below that breakpoint **the class that hid it took the whole answer with it.**

Measured at a real 390×844, coarse pointer, on the bundled sample:

| typed | rows | unexplained (before) | unexplained (after) |
|---|---|---|---|
| `policy` | 4 | **4** | 0 |
| `order` | 20 | **20** | 0 |
| `shopper` | 20 | **20** | 0 |
| `주문` | 11 | 0 | 0 |
| `ㅈㅂㄱㄴ` | 2 | 0 | 0 |

A reason that **carries a mark** now drops to a line of its own under the name. Rows grow 44 → 58px, and only the rows that need it do — the Korean-name matches stay 44. A reason with **no** mark is the summary standing in as context, and the name above it is already the answer, so giving that a second line would double every phone row to repeat something the row is not there for; it stays a `md`-and-up column.

From `md` up nothing moved: still one fixed-width column with one text start line (re-measured at 1440 — `reasonStartLines: 1`, rows 40px).

## The same measurement caught a clipped control

The project filter chip's scroller was a hardcoded `height: 24` around a chip that the touch floor makes **44px** tall. `overflow-x: auto` clips the other axis too, so the chip lost its bottom 20px and the finger got 24. The box now reads `--control-h-sm` — 28 on a mouse, 44 under a coarse pointer.

**No existing gate could have caught it**, because the chip's own rect was already 44 and every case in `touch-target-contract.spec.ts` reads the target's rect. The new case asks the document what is actually at the chip's top and bottom edge:

```
before: chip 44px, scroller 24px, clipped by 20, bottom edge does not hit the chip
after:  chip 44px, scroller 44px, clipped by 0,  both edges hit the chip
```

## Test plan
- `match-reason.test.ts` — `reasonLineClass`: an evidence reason is not hidden below `md`; a context reason is; both are the same fixed-width column from `md` up and neither uses `max-w-`.
- `tests/e2e/touch-target-contract.spec.ts` — a new 390 coarse case. **Gate-probed**: restoring the `height: 24` turns it red with "조상이 잘라내고 있다"; putting the token back turns it green.
- Ran: `eslint --max-warnings 0` on the four changed files · `vitest run src/widgets/global-search` (82 tests) · `pnpm source:language` · `pnpm knip` · `pnpm tsc --noEmit` · `pnpm test:contracts` (272 files, 2729 tests) · `pnpm docs:check` · the whole `touch-target-contract` spec on a port of its own (16 passed).
- Live at 390×844 and 1440×900 against the running build; both tables above are that measurement.


wlsdks:

PR #1743, opened 2026-09-19
fix(map): the chrome may not eat the map on a narrow window

wlsdks:

## Summary
- **The map shrank to a speck on a narrow window.** The camera's side insets are absolute pixels — `--map-safe-inset-left: 350` for the INDEX panel, `--map-safe-inset-right: 120` for the tool rail — and never scaled with the frame. Measured on the sample map: an 820-wide canvas reserved 470 of it (57%) and the graph spanned 31% of the width; a 390-wide canvas reserved more than its whole width, so `effW` collapsed to `Math.max(1, …)` and only `cameraScaleMin` (0.24) kept a frame — the map sat at its floor zoom with nodes spanning 118 px of 390.
- **The fit now reserves at most half of each axis** (`MAX_FIT_CHROME_SHARE`, `clampFitInsets`), shrinking both sides in proportion, and **never below what a panel measurably covers**: `cameraTokens` now passes the measured obstacle along with the tokens, so narrowing the reservation can never slide the graph under an open panel.
- Every desktop width is untouched — their chrome asks for 39% of the canvas, and the widest figure any existing test pins is 46%.

| canvas | drawn width before | after | scale before → after |
|---|---|---|---|
| 1512 · 1280 · 1024 | 37% · 44% · 38% | unchanged | unchanged |
| 820 | 31% | **37%** | 0.522 → 0.615 |
| 640 | 19% | **37%** | 0.252 → 0.48 |
| 390 | 30% | **37%** | 0.24 (floor) → 0.293 |

## Test plan
- [x] RED → GREEN e2e `tests/e2e/map-narrow-viewport-fill.spec.ts`: at 390 the scale is off the floor and the graph uses over 35% of the canvas; at 820 the same, **and its leftmost node stays right of the open panel**; at 1280 nothing moved. RED on the original source (`scale 0.24`).
- [x] Unit `clampFitInsets`: a desktop reservation is left alone; both sides shrink in proportion; the measured floor is never crossed; a floor wider than the ceiling wins; a zero extent changes nothing. 48 pass.
- [x] Siblings: short-viewport fit, viewport reframe, expand-all, smoke, 3D return, galaxy transition, spotlight clearance, edge hover card — 21 passed.


wlsdks:

PR #1744, opened 2026-09-19
test(acp): the work harness drives the request it advertises

wlsdks:
## Summary

This is test infrastructure, not a panel change — but it is the harness every ACP
round is measured through, and it was lying about what it sent.

### The mode drove the wrong tool

`permissionKind: 'ontology-patch'` titled its tool call `patch_concept` but sent
the **wiki** tool name in the correlated MCP payload:

```js
rawInput: { server: "atlas-vault", tool: "write_wiki_file", arguments: rawInput }
```

`toPermissionRequest` derives the tool from that correlation, not from the title.
So the card received a `write_wiki_file` request and classified it through the
**generic** branch of `buildOntologyChangeSet`.

| | before | after |
|---|---|---|
| scope line | `Change your folder · capabilities/task-review · 2 requested fields` | `Update concept · capabilities/task-review · 1 requested field` |
| headline | `Changes 2 fields in task-review` | `Changes the body of task-review` |

The "2 fields" counted the concept's own `slug`. The mode existed and drove the
wrong thing, which is worse than not existing — **every assertion written against
it passed for the wrong reason.**

### The body was accepted and dropped

`permissionText` is in the signature, read only on the wiki branch, silently
ignored on the other. Measured when a probe written to grow the review card until
it scrolled produced a card of **exactly the same height every time** and proved
nothing about the invariant it was written to test.

### The door was narrower than the room

`openLibraryWorkScenario` declared a hand-picked two of the harness's options, so
every other one was unavailable through it without saying so — the same silence.
It takes what the harness takes now.

## Test plan

`pnpm checks:changed -- --run` — 3 checks green: the new Playwright spec,
`source:language`, `tsc --noEmit`. `task-review-permission-geometry.spec.ts`, the
one existing caller of this mode, still passes — it was not depending on the
wrong tool.

**Two deliberate REDs, one per fault, each hitting only its own case:**

| change | fails |
|---|---|
| hardcode the tool back to the wiki one | the patch case |
| drop `permissionText` on the patch branch | the body case |
| either | the wiki case stays green |

⚠️ The assertions are about **what the card receives**, not how it draws it. A
harness is a claim about the wire, and nothing else in this suite checked that
claim.


wlsdks:

PR #1745, opened 2026-09-19
fix(a11y): every INDEX tree row states its level and its place

wlsdks:

## Summary
- **The INDEX tree does not say its own shape.** It is a `role="tree"` whose rows are siblings in the DOM — `0` nested `role="group"` — and the hierarchy is drawn with `marginLeft: depth * 16`. WAI-ARIA then asks each `treeitem` for `aria-level`, and for `aria-posinset`/`aria-setsize` among its siblings. Measured on the live sample map: **10 rows, 0 groups, 0 levels**, so a screen reader announces the project and its nine domains as ten peers; expanding a domain adds its capabilities to that same flat list.
- The depth was already known (it draws the indent), so it now reaches ARIA. Both renderers of the row pass the sibling position: the INDEX tree and the realm ledger.

## Test plan
- [x] RED → GREEN component test in `TopologyIndexPanel.test.tsx`: root at level 1, a domain at 2, an expanded capability at 3, each position inside its set, and no row left without a level. Fails on the original source (`expected null to be '1'`).
- [x] RED → GREEN e2e `tests/e2e/map-index-tree-levels.spec.ts` on the live panel: every row carries a level, the roots are all level 2 under one level-1 project, and expanding a domain produces level-3 rows rather than flattening them.
- [x] `pnpm checks:changed -- --run`: 14/14, including the full contract suite (2,729 tests).
- [x] Siblings: `index-panel-tidy`, `map-narrow-viewport-fill`, `guided-tour` — passed. `map-smoke`'s "opening the doc and going back" is a pre-existing dev-server flake: it fails the same way with this change reverted, and passes on retry.


wlsdks:

PR #1746, opened 2026-09-19
design(map): the relation panel links by name, and a path names both of its ends

wlsdks:
Branches from `main`, independent of the other open map PRs. **Two small map-readability fixes in one pull request**, deliberately: the shared landing queue is ~30 deep and nothing has merged for hours, so a second pull request would have cost more than it explained. Each has its own commit, test and change record.

## 1 — The relation panel links to its document by name, not by arrow
- Clicking a relation opens the edge panel, whose last row linked to the declaring document as **`storefront.md → Open doc`** — an arrow between a file name and an action, inside a `Link` that navigates *within* the app. `.claude/rules/forbidden.md` keeps arrows for path, order, causality or an external destination, and the label-decoration gate says the same in its own failure message.
- The label is the document's name now; the action moves into `aria-label`, so a screen reader still hears it.
- **Why the gate missed it:** its two JSX scans look for an arrow immediately followed by a tag, or an arrow alone in an element. This one sat between two expressions. A measured sweep of `src` and `app` for that shape found 8 — 7 legitimate data arrows inside template literals (`${from} → ${to}`) and this one. No repo-wide pattern was added: it cannot tell a data arrow from a decorative one, and a first attempt produced 6 false positives from comments and from `=>` inside JSX attributes. The blind spot is written down at the fix.

## 2 — A path names both of its ends
- Asking how one concept reaches another opens the path lens: its nodes keep full ink and everything else sinks. Their labels still ranked as ordinary concepts, so an endpoint inside an expanded domain disc lost its slot to that domain (priority 3) and the project (2). Measured on `mode=path` from the exchange-request capability to the stock-reservation one: both ends drawn above 0.89 alpha, **the origin named and the destination anonymous**.
- A lens's own nodes now take the top label band — where the constellation lens already put its members. The rule moved into `resolveLabelPriority` beside the other rankings instead of sitting inline in the draw, so it is testable without a canvas.

## Test plan
- [x] RED → GREEN component test (`OntologyMapEdgePanel.test.tsx`): the link's text is the document name with no arrow glyph, its accessible name carries both the name and the open-document label, href unchanged.
- [x] RED → GREEN e2e (`map-path-lens-names-both-ends.spec.ts`): on a real `mode=path` URL both ends are drawn at full ink **and** both are named.
- [x] Unit (`label-layout.test.ts`): a lens subject ranks 0, above a domain and the project, and every other ranking is unchanged. 36 pass.
- [x] `pnpm checks:changed -- --run`: 12/12 on each change, the first including the full contract suite (2,729 tests).
- [x] Siblings: path lens sink and frame, spotlight clearance, ego dim labels, galaxy transition, expand-all, edge hover card, map smoke — all passed.
- [x] Sweep after the first fix: 0 mid-expression arrows left in JSX text across `src` and `app`.


wlsdks:

PR #1749, opened 2026-09-19
fix(map): discarding a walked trail asks twice

wlsdks:

## Summary
- **One press erased a walk.** The trail popover's footer button and the chip's ✕ both called `onClear` directly. Measured on a three-step walk: pressing it took the trail from 3 rows to 0, removed the chip entirely, and raised **no confirm, no undo button and no status message**. The trail is discarded rather than archived and is not in the URL, so nothing can rebuild it.
- **The same popover already knew better.** A few pixels away, "clear every past walk" arms first and says `Press again to discard`, releasing after four seconds. This makes the session trail's clear behave the same way, so one surface does not hold two destructive controls with two different rules.
- The armed state belongs to the popover, not the button: arming from the ✕ shows the confirm wording on the footer as well.

## Test plan
- [x] RED → GREEN component tests (`TopologyTrailChip.test.tsx`): the first press does not clear and the label changes; the second clears; the ✕ shares the same armed state; the arming releases after four seconds. The two previous tests that asserted a single press clears are replaced by these, which is the behaviour change.
- [x] `pnpm checks:changed -- --run`: 15/15, including the full contract suite (2,729 tests), `map-viewport-reframe`, `ontology-ui` and the desktop check.
- [x] Siblings: `map-smoke`, `map-expand-all` — 7 passed.

The trail popover's step list (relation word per step, "now here" / "N steps back"), its lens (walked nodes numbered and lit, the rest sunk), and the arrow-key walk's dead-end notice all read correctly. A shape I took for a stray empty box inside the popover was the surface itself: sampled at the suspected pixel it is 1 RGB step from the clean surface, contrast 1.01.


wlsdks:

PR #1751, opened 2026-09-19
fix(a11y): the constellation chip expands a named region instead of promising a dialog

wlsdks:

## Summary
- The map's saved-constellations chip carried **`aria-haspopup="dialog"`** while what opened was an unnamed `div`. Measured on the live map with the chip open: **0** elements with `role="dialog"`, **no `aria-controls`** on the trigger, **no accessible name** on the panel, and focus left on `body`. A screen-reader user heard "expanded" and had nothing to move to or announce.
- **It is a disclosure, not a dialog.** Giving it `role="dialog"` would be a hand-assembled modal — `shared/ui/dialog.tsx` owns scrim, trap, Escape, restore and scroll lock, and `dialog-adoption-ratchet` refuses new ones (it caught my first attempt, correctly). It would also be a lie: there is no scrim and no focus trap, focus stays on the chip, and the map behind stays live. So the chip drops the dialog claim, keeps `aria-expanded`, and points at the panel with `aria-controls`; the panel is a `section` named by its own heading.
- The interaction is untouched: Escape closes, an outside press closes, focus never moves.

## Test plan
- [x] RED → GREEN component test: before the click there is no named region; after it, the region exists, carries an id, and the chip's `aria-controls` matches it; the chip never claims `aria-haspopup="dialog"`.
- [x] RED → GREEN e2e `tests/e2e/map-constellation-region.spec.ts` on the live map, including `0` dialogs in the page, the chip keeping focus, and Escape closing it. Verified red against the unmodified source (`element(s) not found`).
- [x] `pnpm checks:changed -- --run`: 13/13, including the full contract suite (2,729 tests) and the dialog adoption ratchet.
- [x] Siblings: `responsive-overflow-audit` and `map-smoke` — 37 passed; the one failure is `map-smoke`'s known dev-server flake ("opening the doc and going back"), which passes on retry and fails the same way with this change reverted.

The constellation panel's own content is honest on the sample vault: a disabled **만들기** and the line "open your ontology folder to save a constellation". Escape, outside-press and focus behaviour were already correct before this change.


wlsdks:

PR #1752, opened 2026-09-19
fix(acp): the empty conversation stops inviting a blank start when the screen already wrote one

wlsdks:
## Summary

Measured across the docks with the `.qa-scratch/acp/` harness at dev :3171, ko + en. The **Insights** dock, at rest:

- the composer already held a question the screen wrote — *"Explain this Analysis tab from the current ontology evidence only"* — with a `이 화면이 붙인 지시 3줄` disclosure folding three instruction lines under it;
- the transcript above it, 470px of empty panel and by far the largest region on screen, read `이 폴더에 대해 무엇이든 물어보세요.` — "Ask anything about this folder."

The panel invited a blank start while the start was already written, and nothing said the staged request was there to be edited or sent. That sentence was written for the Library dock, which seats nothing; the seated docks (Insights, Architecture, topology meaning) inherited it.

A seated dock now says what is waiting:

| dock | before | after |
|---|---|---|
| library (seats nothing) | 이 폴더에 대해 무엇이든 물어보세요. | unchanged |
| insights (seated) | 이 폴더에 대해 무엇이든 물어보세요. | 이 화면이 질문을 하나 써 뒀어요. 아래에서 고치거나 그대로 보내세요. |

The condition is `seenPrefillNonce !== null && draft.trim().length > 0` — the seat **and** the draft. A seat the person cleared really is a blank start again, and the open invitation goes back with them; a draft they typed themselves was never staged by anything.

## Test plan

- `AcpChatPanel.test.tsx` — 136 pass, 3 new: a waiting seated request switches the empty state; clearing the box switches it back; a dock that seats nothing keeps the open invitation. The assertions check `data-acp-empty` and the message key, because this harness echoes keys rather than sentences — and because a gate should not pin prose.
- Rendered evidence, both docks, both locales: `.claude/shots-2026-09-20/acp/r19-{insights,library}-{ko,en}.png` (before) and `r19-after-…` (after).

## Limits

The probe could not open the **architecture** and **topology meaning** docks — both need a node selected first and this probe makes no selection — so the change is measured on the Insights dock only. Both are seated docks and take the same branch; neither is shot here.


wlsdks:

PR #1753, opened 2026-09-19
fix(map): the review chip reports its state and closes what it opened

wlsdks:

## Summary
- `HomePage` states the rule beside the replay control: a control may wear the active tone **and `aria-pressed`** for exactly as long as the thing it names is on. The meaning-review chip wore the tone with **no `aria-pressed`**, and its handler (`openMeaningWorkbench`) only ever opened.
- Measured on the live map with the panel open: **three presses in a row** left the panel present, the map's canvas at **928px** (from 1448), and no pressed state on the chip. A lit control ignored every press, and a screen-reader user was told nothing about the state the colour was showing.
- The chip now reports `aria-pressed` and the press closes what it opened. The panel keeps its own close button; this is the second way in, matching the spotlight chip a few slots away, which already pairs the tone with `aria-pressed`.

## Test plan
- [x] RED → GREEN e2e `tests/e2e/map-review-chip-toggles.spec.ts` on the live map: `aria-pressed` starts false, turns true with the panel open, the second press closes the panel and returns the state, and the canvas goes back over 1200px. RED before (`Expected "false", Received ""`).
- [x] Siblings: `map-constellation-region`, `map-smoke` — 5 passed.

Opening the review panel does not stack surfaces: the node detail panel leaves the DOM rather than sitting behind it, the canvas shrinks instead of being covered, and the map reports no obstacle inset because nothing overlaps it. The panel's content is honest on the sample vault — it says an installed app with a connected agent is needed to analyse.


wlsdks:

PR #1754, opened 2026-09-19
feat(acp): a command in an answer can be taken whole, however narrow the panel

wlsdks:
## Summary

Measured with the `.qa-scratch/acp/` harness at dev :3171, at the dock's own documented floor (`CHAT_WIDTH_MIN`, answer column 260px):

```
pre   pnpm atlas compile --page wiki/architecture.md --force --locale ko
      clientWidth 260 · scrollWidth 456 · hidden 196px (43%)
      overflowX "auto" · scrollbarHeight 2 · edgeMarks [] · copyButton false
```

**43% of the command was unreachable.** The only affordance was a 2px scrollbar — nothing a trackpad offers to grab and nothing an eye reads as an affordance — and there was no copy control anywhere. The one actionable line in the answer could be neither read nor taken.

Every code block in the conversation now carries a copy chip:

```
      copyButton { text "Copy" · 40x24 · ownsPoint true }
```

`ownsPoint` is an `elementFromPoint` check, not geometry: nothing of the block is drawn over the chip.

## Three choices, stated

- **Copy rather than wrap.** Scrolling stays — that is how the rest gets *read*. Copying is how a command gets *used*, which is why it is in the answer. Wrapping would make it readable and lie about it: a wrapped shell command looks like two lines and someone retyping it types the break.
- **Always drawn, not on hover.** There is no hover on a touch screen, and this repository already forbids discovery that exists only under a pointer.
- **Always drawn, not only when it overflows.** An affordance that appears and disappears with the panel width teaches people it might not be there.
- **Under the block, not over its corner.** At 260px a corner chip covers the command it is offering.

It copies the block's rendered text, so what lands on the clipboard is exactly what is shown — not the markdown fence around it.

## Test plan

- `AcpChatPanel.test.tsx` — 135 pass, 2 new: a command is copied whole through a stubbed clipboard and the chip reads back `Copied`; a GFM table gets no chip, because a table is something to read rather than run.
- Deliberate RED: drop the `pre` entry from `chatMarkdownComponents` → the copy case fails, the table case still passes (it is the negative); restored, 135 pass.
- Rendered evidence at 320 / 380 / 520: `.claude/shots-2026-09-20/acp/r20-{before,after}-code-{320,380,520}.png`, plus the full panel sweep in `.claude/shots-2026-09-20/acp/r20/`.

## Not in this change

The edge fade that says *there is more to the right* is a separate open PR of mine from an earlier round; this one is about getting the whole thing, not about marking the cut. The two are complementary and do not touch the same code.


wlsdks:

PR #1755, opened 2026-09-19
fix(a11y): closing the shortcuts sheet returns the keyboard to its button

wlsdks:

## Summary
- **Closing the shortcuts sheet dropped the keyboard at the top of the content.** The sheet records `document.activeElement` on open, but the `?` button unmounts at that moment (raising the sheet turns off its render condition), so the record is empty and the close falls back to `main#main`. Measured on the live map: focus the `?` button, press Escape, and focus lands on `main` **while the button is back in the page**.
- The sheet now takes `returnFocusSelector`, looks the control up after it remounts, and keeps the old fallback when there is none — the same fix `AnalysisWorkbench` and `ArchitectureAgentDock` already use for this exact unmount.
- **Also: the map probe's `alpha` no longer promises what it cannot deliver.** Its doc-block said "the alpha the last frame drew this node at ... whether the thing is visible is this number", but it comes from the pass *before* the lens sink and the growth-replay ramp. That promise cost three wrong readings in one day (a path whose surroundings were plainly sunk reported 0 dimmed nodes; a replay that grew from 91 to 1,134 lit pixels reported one unchanging state). The comment now says what it is and points at canvas-pixel sampling, which is what `map-path-lens-sink.spec.ts` already does.

## Why the probe is documented rather than fixed
I tried recording `ctx.globalAlpha` at the node body draw and **backed it out**: nodes are drawn in more than one pass, so a single hook produced a worse instrument than the blind one — it reported the darkest nodes on screen (median 43 of 765) as `1.00` and the sunk ones (median 195) as `0.30`. Making it faithful means understanding the multi-pass draw, which is its own piece of work, not a loop round. Shipping a wrong instrument is worse than a documented blind one.

## Test plan
- [x] RED → GREEN e2e `tests/e2e/map-shortcuts-focus-return.spec.ts`: open with the `?` button, Escape, the button is visible and focused. RED before (`Expected: focused, Received: inactive`, 33 retries).
- [x] `pnpm checks:changed -- --run`: 13/13, including the full contract suite (2,729 tests), the sheet's own 13 unit tests, `map-viewport-reframe` and `ontology-ui`.

The growth replay works — it grows from 91 to 1,134 lit pixels over 6.4s and returns its button to rest by itself. The shortcuts dialog is otherwise correct: modal, named, focus moves in, Escape closes. Every other map chrome control that wears an active tone reports it (`aria-pressed` on the toggles, `aria-expanded` on the popups); the one exception is already fixed in #1753.


wlsdks:

PR #1756, opened 2026-09-19
fix(acp): the error that cannot retry opens the screen it names

wlsdks:
## Summary

Measured with the `.qa-scratch/acp/` harness at dev :3171, both locales. The **launch** error card, rendered for the first time this round:

```
도구를 띄우지 못했어요
왼쪽 「에이전트」 화면에서 이 도구가 어떤 상태인지 확인해 주세요.
자세히 · spawn npx ENOENT
doors: [ button "연결 점검" 81x32 ]
```

**The card names a screen and offers no way to it.** Its one control is the connection check — which the panel's own comment calls the row's *second* action, the one meant to sit after a retry. On this card the second action is the only action, and it is not the one the sentence asks for.

The retry was added to this card in August precisely to close that shape: five of six kinds ended in "press New chat" and New chat was a pencil glyph two hundred pixels away (*"if this is normal I still would not know what to do"*). `launch` was excluded on purpose and the reasoning is right — another attempt at a start we already know cannot work teaches people to distrust the button. But excluding the retry left the dead end in place for the one kind that most needs a way out.

The direction was also a guess: the rail that holds Agents is hidden below `lg`, so "on the left" is not true at every width the dock lives at.

After:

```
도구를 띄우지 못했어요
이 도구가 지금 어떤 상태인지 「에이전트」에서 볼 수 있어요. 아래 버튼으로 바로 열어요.
doors: [ a "에이전트 열기" → /ko/agents/ 94x32 ownsPoint:true , button "연결 점검" ]
```

The destination is the attention winner; the connection check stays second; the retry stays absent.

## Harness

`installLibraryWorkHarness` gained `startError`, which makes `acp_start` reject with a given message. The panel classifies a failing start through `readAcpTrouble`, so the message picks the card. **Without it none of the seven error cards had ever been rendered in a test** — which is why this one went unexamined.

## Test plan

- `tests/e2e/agent-trouble-next-step.spec.ts` — 3 new cases, all pass: the launch card carries a door to `/agents` and no retry; a network failure keeps its retry and gets no door; pressing the door actually arrives at Agents. `ownsPoint` is an `elementFromPoint` check, not geometry.
- Deliberate RED: restore `trouble?.kind === 'launch' ? null :` → 2 of the 3 fail (the network case is the negative and stays green); restored, 3 pass.
- This is an end-to-end gate rather than a unit one because the card is reached through a failing start, which is a bridge fact and not a prop. I tried the unit route first and the error card never rendered from a prompt-level error for this kind; rather than guess at why, I gated it where the behaviour is real.
- `pnpm checks:changed -- --run` — 9/9 pass.
- Rendered evidence: `.claude/shots-2026-09-20/acp/r21-{before,after}-{launch,network}-{ko,en}.png`.

## Note

`hover-axis-adoption-ratchet` caught the first version: I had copied the retry chip's literal `hover:bg-[…]`, which would have raised the hand-written-hover count. The link takes `hoverSurface: 'lift'` from the control axis instead; the retry beside it predates that axis and stays on the ratchet.


wlsdks:

PR #1757, opened 2026-09-19
design(projects): the project screens speak in one voice

wlsdks:
## Summary

Measured across `messages/ko.json` on 2026-09-20: the project detail screen writes 16 strings in the soft `-어요` register and 4 in the formal `-ㅂ니다`, and the four outliers sit exactly where a reader meets both at once.

| where | before | register |
|---|---|---|
| right rail, connected card | 아직 다른 프로젝트와 연결되지 않았습니다. | formal |
| right rail, its hint | …여기 나타납니다. | formal |
| left column, domain zone | 아직 도메인이 없어요 / …나타나요. | soft |

Those two are both empty states, drawn in the same row of the same page. `bodyEmptyHint` and `handoffDesc`, the card directly under that rail, were the other two.

On the list the opening sentence is soft and the note printed two lines under it was formal, so the intro block contradicted itself in four lines.

Six strings now speak the voice their own screen already speaks. The list's sample note is rewritten rather than converted word for word: its verb has no natural soft form that a reader would not also read as an instruction, so the sentence says the same fact a different way.

**This is deliberately not a house-style sweep.** Across the whole product the two registers run 713 to 787, so there is no single voice to impose and most strings elsewhere are left alone. The rule is narrow, and it is the whole of the change: a string may not contradict itself, and on a screen with a clear majority the few outliers follow it. `messages/en.json` is untouched, because English carries no register.

## Test plan

- [x] `pnpm checks:changed -- --run`: `test:desktop:check` and `test:i18n:messages` green, twice
- [x] Browser, `/ko/project/fallback/?slug=storefront` at 1512: no leaf element on the page ends in the formal register; the connected card and the domain zone now read alike
- [x] Browser, `/ko/projects/`: the same probe returns nothing formal, and the intro block reads in one voice
- [x] No test pins any changed string. `작성된 본문이 없습니다.` does appear in a test, but that is `fullDetailA1.body.empty`, a different key with its own inline mock


wlsdks:
Correcting my own test plan before this lands: one of the six strings is not drawn anywhere.

`projectPages.selector.nextSlotHumanLead` has no consumer. Nothing in `src/` or `app/` calls it, so the register fix on that line changes nothing a reader will see. The other five are live and were measured in the browser, which is exactly why the browser probe did not catch this one: a string that never renders cannot fail a render check.

The change is still correct and still worth landing. Read the summary as five rendered strings plus one edit with no effect.

It also turned up something larger, which I will raise separately: 23 of the 39 strings in `projectPages.selector` have no consumer, together with `buildRecentActivityRows` and its six tests. They are the remains of the full-width card that the 2026-09-15 compact-index decision replaced — the fact strip, the "next project" slot with its CLI and agent lines, and the recent-activity strip. More than half of that screen's copy describes sections it no longer draws, which is how I came to edit a dead line in the first place.

PR #1758, opened 2026-09-19
design(acp): the permission button that writes stops looking like the one that refuses

wlsdks:
## Summary

Measured with the `.qa-scratch/acp/` harness, both locales, computed styles rather than sight. With the meaning review open the decision row draws four buttons of exactly **210×32**, and two came back byte-identical in paint:

```
task-review-correct    거절 후 수정 / Reject and edit   bg rgba(255,255,255,0.02)  border rgba(255,255,255,0.1)
task-review-defer      대기 유지    / Keep pending      bg rgba(0,0,0,0)           border rgba(0,0,0,0)
acp-permission-reject  안 할래요    / No thanks         bg rgba(0,0,0,0)           border rgba(0,0,0,0)
acp-permission-allow   이번만 허용  / Allow once        bg rgba(255,255,255,0.02)  border rgba(255,255,255,0.1)

identical paint pairs: [correct ↔ allow] , [defer ↔ reject]
```

**One of that first pair refuses the write. The other performs it, on the person's own files, and nothing on this card undoes it.** Four identical rectangles in a 2×2 grid and the only one that touches the folder looked exactly like a refusal.

The two-button branch had always given allow the filled `primary`; the review branch had quietly demoted it to `outline`, which is the variant `Reject and edit` wears. Promoting it back leaves exactly one filled control in the row, and it is the irreversible one:

```
acp-permission-allow   이번만 허용   bg rgb(94,106,210)  ink rgb(255,255,255)
identical paint pairs: [defer ↔ reject]
```

The surviving twin is honest: `Keep pending` and `No thanks` both mean nothing is written now.

## Test plan

- `tests/e2e/permission-decision-paint.spec.ts` — 2 new cases, both pass. The gate asserts the **pair**, not a colour: the writing decision must not share background, border, ink and weight with any decision that writes nothing. It survives a re-theme and fails the moment the two are flattened together again.
- Deliberate RED: put `variant="outline"` back → the paint case fails, the focus case stays green (it is about a different property); restored, 2 pass.
- `pnpm checks:changed -- --run` — 7/7 pass.
- Rendered evidence: `.claude/shots-2026-09-20/acp/r22-{before,after}-decisions-{ko,en}.png`.

## A claim I had to correct mid-round

I first wrote the second gate as *"the row still opens with the keyboard on reject"*, because that is what the panel's unit test asserts. **It failed.** Measured in the real browser, focus opens on `task-review-heading` — the card's own "Current task" heading — not on any decision.

That is stronger than parking it on reject, and it means the fill does not put the keyboard anywhere dangerous; but it is not what I assumed, so the gate now asserts the property that is actually true and actually matters: **focus does not open on the button that writes.** It holds whichever of the two focus strategies the card uses.


wlsdks:

PR #1759, opened 2026-09-19
fix(map): a realm calls its root one name

wlsdks:

## Summary
- Entering a realm (`?realm=domain:payment`) names its root in **four places at once**: the chip, the chip's title, the node on the canvas, and the ledger's header. Measured on the sample map: the first three read the vault's per-locale display name, and the header read the canonical `title` — so one screen called the same domain by two names in two languages.
- The ledger's own boundary rows and subtree were already localized (`부분 취소 → 환불 처리`, `네이버페이 연동`); only `realmLedgerModel.rootTitle` reached past `display`. It now uses the `display ?? title` idiom the rest of `HomePage` uses.

## Test plan
- [x] RED → GREEN e2e `tests/e2e/map-realm-one-name.spec.ts`: inside a realm, the chip title and the ledger header both equal the label the canvas draws. RED before (`Expected "결제", Received "Payments"`).
- [x] Siblings: `map-smoke`, `index-panel-tidy` — run below.

The realm itself reads well: the URL carries it (`?realm=`), the ledger states the census (`elements 9 · capabilities 6 · depth 2`), it counts the relations crossing the boundary (7) and offers a jump to each one's own realm, it offers saving the group as a folder, and two separate controls return to the whole map.

## A gate note
Two failures before this was green were mine, not the code's: the first gate run started while the dev server was still rebuilding after the edit (10 specs failed, all passing on a rerun), and the source/docs language gates caught Korean glyphs I had put in a code comment and a spec doc-block.


wlsdks:

PR #1760, opened 2026-09-19
design(acp): the permission card draws a status row only where there is a status

wlsdks:
## Summary

Measured with the `.qa-scratch/acp/` harness at dev :3171. The status grid above the permission decision, before:

```
box: y 642 · 428x41
의미        알 수 없음
코드 검증   알 수 없음
병합        알 수 없음
배포        알 수 없음

distinctValues: ["알 수 없음"]
```

**Four label/value pairs, one distinct value.** And three of them could never say anything else — `code`, `merge` and `deployment` were written as the literal `'unknown'` key:

```tsx
{t(`taskReview.authority.${authority === 'meaning' ? meaningAuthority : 'unknown'}`)}
```

So they were not "we do not know yet"; they were "this screen cannot know, and never could". Only `meaning` moves, between unknown / pending / accepted.

The claim those three rows were standing in for is already a sentence on the same card, unfolded, measured in the same run:

```
이번만 허용은 이 쓰기만 진행합니다. 의미 승인이나 코드 검증, 병합, 배포를 뜻하지 않습니다.
```

That names all four and says what allowing does **not** grant. "Unknown" is a different and weaker claim — it says the state is unavailable, when the point is that allowing here does not touch it.

After:

```
box: y 660 · 428x23
의미        알 수 없음

distinctValues: ["알 수 없음"]
```

41px → 23px, directly above the decision buttons, where attention is scarcest.

## Test plan

- `AcpPermissionCard.test.tsx` — 36 pass. The case that used to pin *"all four authorities are unknown"* now states the rule: only the row that moves is drawn, the other three are absent, and the standing sentence still names all four words so nothing was lost.
- Deliberate RED: restore the four-row map → that case fails; restored, 36 pass.
- `pnpm checks:changed -- --run` — 6/6 pass.
- Rendered evidence: `.claude/shots-2026-09-20/acp/r23-{before,after}-authority-{ko,en}.png`.

## Note

The test that failed here was a test of mine from an earlier round pinning the defect as if it were the specification. Rewriting it was the change, not collateral.


wlsdks:

PR #1761, opened 2026-09-19
fix(map): the agent chip closes what it opened

wlsdks:
Branches from `main`, independent of the other open map PRs. Same shape as #1753, found by carrying its rule to the next control.

## Summary
- The map's agent chip wears the active tone, reports `aria-expanded`, and called only `openVaultAgent`. Measured **with a folder open** (desktop bridge + fixture vault) at 1512 wide: the first press opened the dock and narrowed the map's canvas **1448 → 1055**, and the second and third presses left `aria-expanded="true"` and the canvas at **1055**. A control that says "expanded" while ignoring the press is not a control.
- It now closes what it opened, using the `closeVaultAgent` that already existed. The dock keeps its own close button; the chip is the second way, and the one a person reaches for after opening it there.

## Test plan
- [x] RED → GREEN e2e `tests/e2e/map-agent-chip-toggles.spec.ts`, run against the desktop runtime with a real folder: open narrows the canvas, the second press returns `aria-expanded` to false **and gives the canvas its full width back**. RED before (`Expected "false", Received "true"`).
- [x] Siblings: `desktop-rail-arrival`, `map-smoke` — run below.

- **A name sweep across the map surfaces** for the canonical title leaking past the reader's display name, after #1759 found one: 17 sites already use `display ?? title`, and the 33 bare `.title` reads are all on objects built with the display name upstream. The trail, the realm tree and the boundary rows all render the localized name.
- Right-clicking empty canvas with no folder open leaves the browser's own menu alone and draws no door that will not open.
- With a folder open, the map draws 5 nodes and names all 5.


wlsdks:

PR #1762, opened 2026-09-19
fix(acp): a tool call wrapped by the MCP envelope still names what it touched

wlsdks:
## Summary

Measured with the `.qa-scratch/acp/` harness at dev :3171, driving a write that gets rejected. Two rows in the same trace:

```
read_source · sources/architecture.docx   완료 / done      target "· sources/architecture.docx"
write_wiki_file                           실패 / failed    target null
```

**The row that failed named nothing.** The read said which file it read; the write said only its tool name.

The adapter reports a call to an MCP server as `{ server, tool, arguments: { … } }` whenever it carries `is_mcp_tool_call`, and both `readToolTargets` and `readToolFallbackTarget` looked only at the top level of `rawInput`. Every key they wanted was one level down.

That was blind in the two places it costs most:

- the wrapped shape is what **this product's own vault server** produces, so it was the common case, not an edge one;
- and the row that lost its subject was **the one that had just failed**. A failure with no subject is the least readable line in a trace.

After:

```
write_wiki_file · outside/unknown.md      실패 / failed
```

`outside/unknown.md` is exactly what that scenario tries to write, so the row now explains its own failure.

`acp-client.ts` already required `server`, `tool` and an object `arguments` together before trusting this shape (`readCorrelatedMcpToolContext`). The same three are required here, so a tool that merely happens to take an `arguments` parameter is left alone.

## Test plan

- `tool-targets.test.ts` — 24 pass, 4 new: the path is read out of the envelope; a vault node is too; a tool with an ordinary `arguments` parameter and no `server`/`tool` beside it is untouched; an envelope whose arguments name nothing still returns null.
- Deliberate RED: drop the unwrap from both readers → the two positive cases fail, the two negatives stay green; restored, 24 pass.
- `pnpm checks:changed -- --run` — 5/5 pass.
- Rendered evidence: `.claude/shots-2026-09-20/acp/r24-{before,after}-failed-{ko,en}.png`.

## A wrong turn I backed out of

I first read the same measurement as *"done and failed are painted identically"* — both came back `rgb(138,143,152)` at 11px. Re-measuring with `fontWeight` showed the failure at **560** against the success's **400**, so the distinction exists and is deliberate; the quiet lane is not supposed to shout. I dropped that and took the gap that survived measurement.


wlsdks:

PR #1763, opened 2026-09-19
fix(map): a name blocked below tries the slot above before it is dropped

wlsdks:

## Summary
- The greedy label placer takes candidates by priority and **drops** any whose box overlaps one already placed. There was no second attempt, however much room sat beside it.
- Measured with a real folder (desktop bridge + fixture vault, 20 concepts) at 1512×982 with every group opened: the map used **49% of the canvas width** and still drew one capability **without its name**, beaten by a neighbour **49px away**, while its own upper slot was clear.
- The frame already computes that upper slot for every node — it is what the existing "blocked below, flip above" pass uses against reserved discs. The placer now takes it as `altBbox` and drops a name only when **both** slots are taken, so the no-overlap rule is unchanged.

## Test plan
- [x] RED → GREEN e2e `tests/e2e/map-label-tries-the-upper-slot.spec.ts`: with every group opened, no drawn concept is missing its name. RED against the unmodified source (**3 unnamed**).
- [x] Unit `label-layout.test.ts`: the fallback places the blocked name in the alternate box and marks it; both slots taken still drops it; no alternate behaves exactly as before; a free first slot is never pre-empted. 38 pass.
- [x] Siblings: spotlight clearance, floor labels, ego dim labels, path lens frame, expand-all, index tree levels — 5 passed.

With a real folder open: the recent tab's count is honest (20 of 20 concepts changed within a day, and pressing it draws all 20 and names all 20); the "documents not on the map" row counts correctly and disappears when there is nothing to tidy (0 → no row, 1 → "1개", 3 → "3개"); canvas fill holds at 48–51% width and 65–70% height across arrival, the recent lens and expand-all, with nothing off-canvas.

## Landing note
`pnpm pr:land 1728` had fallen out of the FIFO line at `refs/atlas/landing-queue` and was only racing for the lock. I merged main into that branch and restarted its lander so it re-registers.


wlsdks:

PR #1764, opened 2026-09-19
design(acp): the compare tab says "there is no before" once, and lets the reason give the cause

wlsdks:
## Summary

Measured with the `.qa-scratch/acp/` harness at dev :3171, Compare tab with no comparison basis. **The same fact, five times on one screen:**

```
y=368  이전 값을 지어내지 않고 현재 요청값만…   ← standing sentence (claims: no trusted snapshot)
y=390  아직 비교할 수 없는 요청 형식입니다.      ← reason (claims: request shape)
y=439  이전 · 확인할 수 없음                    ← per unit
y=514  이전 · 확인할 수 없음                    ← per unit
y=588  이전 · 확인할 수 없음                    ← per unit

host 428x369 · overflowY 54
```

Three copies of a line that never changes, pushed between the three proposed values a person opened this tab to read. And the tab overflowed by 54px carrying them.

**Worse than the repetition, the first two disagree about why.** The standing sentence asserts a missing historical snapshot; the reason says the request shape is not comparable. Only one of those was true, and it was the specific one — the generic sentence was guessing at a cause it does not have.

After:

```
y=368  이전 값을 지어내지 않고, 지금 요청에 담긴 값만 보여 드려요.   ← the promise, which always holds
y=390  아직 비교할 수 없는 요청 형식입니다.                          ← the cause, from the side that knows it
y=439  Review the selected proposal.            제안값
y=496  - Only the exact current request…        제안값
y=552  - Allow once does not accept meaning.    제안값

host 428x369 · overflowY 0
```

Every value still carries its `제안값` / `After` prefix, so each is still marked as proposed rather than current. Nothing about the absence is lost — it is said once, and its reason is said by the code that knows it.

## Test plan

- `AcpPermissionCard.test.tsx` — 37 pass. The compare case now asserts the rule: the sentence appears exactly once, `Before · unavailable` appears nowhere, and every proposed value still carries its prefix. A second case holds both catalogues to the sentence naming no cause it cannot know.
- Deliberate RED: put the per-unit line back → the first case fails; restored, 37 pass.
- Rendered evidence: `.claude/shots-2026-09-20/acp/r25-{before,after}-compare-{ko,en}.png`.

## One thing this does not gate

That the reason line is *drawn* is not asserted in jsdom: this card's fixture reports no `unavailable` reasons, so the list is empty there. I tried it, it failed, and rather than bend the shared fixture I narrowed the gate to what it can prove and left the reason to the rendered evidence — where `unsupported_request` is visible in both locales.


wlsdks:

PR #1769, opened 2026-09-19
design: no Korean sentence hands the reader a choice of particle

wlsdks:
## Summary

Korean picks a particle by whether the preceding syllable ends in a consonant. When the value is interpolated, its ending is unknown at authoring time, ICU cannot choose, and the writer leaves a parenthesised pair behind. **Eighteen strings across the catalogue shipped one to the reader.**

One of them is the line that tells a person a file was **written without asking** — the highest-stakes sentence the chat panel has:

```
acpChat.notice.autoAllowed
  before   {detail}을(를) 묻지 않고 썼습니다. 위키 서식에 맞아서요.
  after    묻지 않고 {detail}에 썼습니다. 위키 서식에 맞아서요.
```

There is no correct particle to substitute, so none is substituted. Each sentence is rephrased into a shape that needs no allomorph — a postposition with a single form, an apposition, or the name moved out of the clause behind a separator:

```
connectors.removeConfirmTitle   {name}을(를) 지울까요?        →  지울까요? · {name}
library.sources.state.compiled  {pages}이(가) 정리했고, …      →  정리한 문서: {pages}. …
library.wiki.originalMissing    {name}을(를) 인용하지만 …      →  근거로 인용한 {name}, …
connectors.rowTransfer          …{destination}과(와) 직접 …    →  …직접 주고받습니다 · {destination}
```

Nothing loses a fact; several read shorter.

## Test plan

- New contract `korean-particle-choice.contract.test.ts` holds the Korean catalogue at zero and is itself held to a planted example, so it cannot pass by having no subjects.
- **The gate found the eighteenth string.** My own scan matched `과(와)` in one order only; the gate's pattern covers both and caught `connectors.rowTransfer`.
- Two entries in the existing `korean-particle-spacing` contract covered sentences that no longer exist. That contract already refuses a stale exemption — *"every entry still earns its place"* — which is how they surfaced; both are removed and its probe now uses an entry that is still live.
- `pnpm checks:changed -- --run` — 5/5 pass.

## What I could not do

**No rendered evidence for the notice itself.** I tried twice to drive `acpChat.notice.autoAllowed` onto the screen — `writeMode: 'auto'`, then again with a wiki page built to satisfy the page contract — and it did not appear either time; the auto-decide path needs more than the harness gives it. Shots of the attempts are in `.claude/shots-2026-09-20/acp/r29-before-notice-{ko,en}.png` and show the panel without the notice.

The defect is in the catalogue, which is the source of what every screen prints, and the fix is verified there. But I did not see this one on a screen and am not claiming to have.


wlsdks:

PR #1771, opened 2026-09-19
fix(acp): the automatic write receipts — reachable, pressable, and the refusal says it refused

wlsdks:
> **Two commits, and that was a mistake of mine.** Round 30's work (reachability + touch floor) and round 31's (the missing refusal branch) both landed on this branch because I reused the worktree without switching branches first. They touch the same element in the same file, so rather than rewrite pushed history I am describing both here. If you would rather review them apart, say so and I will split them.

## 1 · The receipt could not be reached, so nothing had ever measured it

Reaching this notice needs four things at once — the folder set to write automatically, a permission carrying a real `filePath` rather than the MCP envelope, a page that passes the wiki contract, and a native folder root. No spec had assembled them, so the one surface reporting **a write nobody approved** was drawn only by the product.

Assembled here for the first time:

```
entries: ["user","tool-run","tool","tool","notice"]
wiki/architecture.md…  묻지 않고 썼습니다. 위키 서식에 맞아서요.   [열기] [다음엔 묻기]
```

### What that hid

Under `(pointer: coarse)`:

```
열기         16 × 24
다음엔 묻기   43 × 24
```

against the **44px** `--touch-target-min` applied through `.atlas-touch-floor`. A sixteen-pixel target is not reachable by a finger, on the notice where the way back matters most. `touch-target-contract.spec.ts` sweeps real controls against that floor and would have caught it — it never saw these two.

After: both **44 × 44**.

## 2 · Its refusal sibling drew an unrelated sentence

The session pushes an `auto-refused` notice whenever `autoDecide` answers with a rejection, and `use-rounds-runner` answers that way on **every write outside a round's scope**. The transcript's notice chain had no arm for it, so it fell through the final `else` to `notice.gateOff` — *"Atlas cannot ask you before a file outside this folder is touched."*

The opposite of what happened: Atlas had decided, and refused. The subject it blocked (`detail`) was dropped with the sentence.

There was no `notice.autoRefused` string either, and the cause is upstream of both: the panel's own `autoDecide` prop was typed `string | null` while the session it feeds has always taken `string | { reject } | null`. **The case the product emits could not be written down in the component that draws it.** `tsc` is what surfaced that — the first version of the test would not compile.

The prop now matches the session, the branch exists, and the sentence names what was blocked. No doors: unlike the allow receipt there is no page to open and nothing to put back.

## Test plan

- New `tests/e2e/agent-auto-allowed-receipt.spec.ts`, 2 cases: the notice names the page it wrote, offers both doors, stands without a permission card; both doors clear the floor in width and height, read from `--touch-target-min` at runtime rather than copied as `44`.
- `AcpChatPanel.test.tsx` — 134 pass, 1 new: an automatic refusal says it refused and names what it blocked, and is not the gate-off sentence.
- Deliberate RED, both: removing `atlas-touch-floor` fails the floor case and leaves the reachability case green; removing the `auto-refused` arm fails the refusal case.
- Rendered evidence: `.claude/shots-2026-09-20/acp/r30-before-notice-{ko,en}.png`.

## It also closes a gap I left open

[#1769](https://github.com/wlsdks/ontology-atlas/pull/1769) fixes the Korean particle in this exact sentence and shipped saying I could not put the string on a screen. This is the screen; the shot shows it as it reaches a reader today.


wlsdks:

PR #1773, opened 2026-09-19
design(acp): the presentation block card names only reasons it can give

wlsdks:
## Summary

The card that says **why an answer cannot become a presentation** listed ten reasons. Two of them could never be read, because the panel re-checks the same conditions before it draws the card at all:

| reason | the builder returns it when | the panel already requires |
|---|---|---|
| `intent_inactive` | the intent is null, **or** this turn is not the request | `presentationIntent !== null` **and** `currentTurnIsPresentationRequest` |
| `turn_incomplete` | `sessionStatus !== 'ready'` | `status === 'ready'` |

Both sentences were written, translated, reviewed and shipped for a state the surface cannot be in:

```
intent_inactive   발표 요청이 아닙니다.          It is not a presentation request.
turn_incomplete   답변이 아직 끝나지 않았습니다.   The answer is not finished yet.
```

They are gone. Ten reasons, eight of them possible.

### One that looks the same and is not

`no_answer` stays. The panel looks for an agent answer **anywhere in the conversation**; the builder looks for one **inside this turn**. A turn that answered nothing after an earlier turn that did still reaches it. I nearly removed it on the symmetry before checking, and the new test pins that difference so nobody removes it later on the same reasoning.

## Test plan

- New `presentation-block-reach.test.ts`, 2 cases: under the panel's own three guards the builder never answers with either removed reason; and the catalogue carries a sentence for `no_answer` in both languages while carrying neither removed one.
- Deliberate RED: put `intent_inactive` back into `ko.json` → the catalogue case fails; removed again, 2 pass.
- `pnpm checks:changed -- --run` — 7/7 pass.

## Why the test matters more than the deletion

Removing copy only pays if the unreachability is checked. If a guard is relaxed later, the reason becomes reachable and `t()` would print the raw key on screen. The test fails on the reachability first and on the missing sentence second, so the two halves cannot drift apart.

## Scope note

This is the second surface in three rounds found by asking *what can this panel draw that nothing has ever drawn*. Round 30 found two doors below the touch floor that way; this one found copy for states that cannot happen. The presentation feature has more unmeasured surface — the scene view, its evidence strip and its two arrows have no test that references them — and that is a round of its own.


wlsdks:

PR #1774, opened 2026-09-19
test(acp): the composer footer sweep opens the dock once instead of thirty times

wlsdks:
## Summary

**This file broke two sessions' landings tonight.** It opened the dock per case — thirty cases, each a viewport, a seed, a harness, two navigations and three clicks.

```
before   30 cases, 30 boots, 1m18s locally
after     4 cases,  4 boots,   24s locally
```

Locally the old shape was 2.6s a case and fine. In CI it turned **shard 1 of 3 from its usual eleven minutes into thirty**, with cases timing out at exactly 60s and then passing on retry in 2.1s. Every one of those retries is a landing that has to be run again, against a REST quota the fleet is already exhausting — measured at 228/5000 with 21 landers polling while I was writing this.

The dock publishes its width as `--library-agent-chat-width` and the panel is laid out from it, so every question this file asks — truncation, hiding, overflow — is answered by CSS from the box. Setting the variable moves the box: `520 → panel 462`, `320 → panel 262`.

## The verdicts are unchanged, and I checked rather than assumed

Both versions were run against two trees:

| tree | old spec | new spec |
|---|---|---|
| current | 30 passed | 4 passed |
| `shrink-[99]` removed from the runtime name | 30 passed | 4 passed |

Same answers on every width and locale. **Neither version fails on that flip**, so I am not claiming a RED demonstration of the original defect for either — what I can show is that the rewrite did not change what this file concludes.

## What is newly falsifiable

Correctness now depends on the resize taking effect, so each run proves it did: distinct requested widths must produce distinct panel boxes. Stubbing the `setProperty` call turns the sweep red and says what it measured:

```
Error: the dock did not move: 5 widths produced 1 box(es) (262, 262, 262, 262, 262)
```

Without that, a broken resize would leave a file that measures one width nine times and passes for ever — the shape this repository already records as *"a gate with no subjects prints a pass"*.

## Test plan

- `pnpm checks:changed -- --run` — 3/3 pass.
- The equivalence table above, and the stub-the-resize falsification.


wlsdks:

PR #1775, opened 2026-09-19
fix(acp): a session that ends says what the agent printed

wlsdks:
## Summary

Loop round 53 (2026-09-20), priority ①. On the installed app every agent press now fails: the log says `acp session acp-N exited with code Some(0)` within a second, the conversation shows a card blaming a stale app-scoped login, and its retry loops. Outside the app the same account and the same adapter are healthy — `claude -p` answers, and `npx -y @agentclientprotocol/claude-agent-acp@0.70.0` initializes even with `CLAUDE_CONFIG_DIR` set to the app's isolated folder. The cause is inside the app's session path, and it is invisible: the child's stderr is piped to the screen as `acp://stderr` and nowhere else, so a child that dies on arrival takes its own words with it.

**A session that dies on arrival says what it printed.** The stderr pump keeps the child's first three lines (each trimmed and capped at 200 characters), and the thread that waits on the child writes them to the log when the session ends within ten seconds of starting: `acp session <id> said [0]: …`, or `ended within 10s and printed nothing` when it said nothing at all. A session that lives past that window logs nothing, so a working folder's log stays as quiet as before.

The cap is the whole discipline: a child may print a token-bearing URL or a stack that runs for pages, and neither belongs in a log a person may paste into an issue. Three lines of two hundred characters name a cause — "command not found", "invalid API key", "ENOENT" — and no more.

This is the instrument, not the fix. With it, the next round can name why the installed app cannot start a turn.

## Evidence

- Rust unit: `a_logged_line_is_trimmed_and_capped` (trims, caps at 200 + ellipsis); `cargo test --lib` 1 pass, `cargo fmt`, `cargo check` and `cargo clippy` clean for this file.
- Installed app rebuilt from this branch and the failing press repeated; the log now carries the child's own words (see the follow-up comment).
- `pnpm checks:changed -- --run`.

## Test plan

- [x] `cargo test --lib`, `cargo fmt`, `cargo clippy`
- [x] Installed app: press with a dead session and read the log
- [ ] `pnpm checks:changed -- --run`


wlsdks:
Measured with the instrument on the installed app (rebuilt from this branch, trial vault `acp-trial-small`):

- Press 서식 맞는지 확인 → a session starts and exits with code 0 after roughly a minute.
- The new log lines show the child printed **nothing at all** on its error stream; there is no `said [0]:` line and no login or auth text anywhere in the log.
- So the conversation's card ("이 앱 몫으로 남은 옛 로그인이 가로막고 있어요") is not supported by anything the child said. Outside the app the same adapter and account are healthy: `claude -p` answers, and `npx -y @agentclientprotocol/claude-agent-acp@0.70.0` initializes **with `CLAUDE_CONFIG_DIR` set to the app's own isolated folder**.
- A child that prints nothing and exits 0 after a while is what a closed stdin looks like. The next round follows that thread (who drops the session's stdin), not the keychain.

wlsdks:

PR #1776, opened 2026-09-19
fix(acp): the presentation's arrows can be pressed by a finger, and are finally tested

wlsdks:
## Summary

Checked across the repository — every control on this panel is referenced by some test except the two arrows:

```
acp-presentation-citation  → AcpChatPanel.test.tsx, AcpPresentationPanel.test.tsx
acp-presentation-ask       → AcpChatPanel.test.tsx
acp-presentation-open-map  → both
acp-presentation-previous  → NONE
acp-presentation-next      → NONE
```

They are the most pressed controls the panel has. Walking a presentation is Next, Next, Next.

### They were also the small step

`--control-h-sm` is **28px**, and the `button` shape carries no coarse promotion — only `chip`, `row`, `pill` and `segment` do. So all three footer controls sat **16px under** the 44px `--touch-target-min` on a touch screen.

`touch-target-contract.spec.ts` sweeps real controls against that floor and would have caught it. It never saw these: the presentation needs a whole ready trace to appear and no end-to-end spec can reach it — the same blind spot that hid the auto-allowed receipt's two doors in [#1771](https://github.com/wlsdks/ontology-atlas/pull/1771).

## Test plan

- `AcpPresentationPanel.test.tsx` — 6 pass, 4 new: no way back from the first scene; one scene at a time; the last scene **ends** the walk rather than pretending there is another (Finish closes instead of stepping past the end); and all three footer controls carry the coarse floor.
- Deliberate RED: make `next` always step forward and drop one floor class → exactly those two cases fail; restored, 6 pass.
- `pnpm checks:changed -- --run` — 6/6 pass.

## Two things stated honestly

- **The floor is checked as a class, not as pixels.** jsdom cannot measure a box, and this surface is not reachable end-to-end, so I could not repeat #1771's `(pointer: coarse)` measurement here. The 28-vs-44 figure is read from `--control-h-sm` and the shape table in `control-class.ts`, not measured on a screen.
- **I overstated this surface last round.** [#1773](https://github.com/wlsdks/ontology-atlas/pull/1773) said "the scene view, its evidence strip and its two arrows have no test that references them". Only the arrows were true; the scene view and the citations were already covered by `AcpPresentationPanel.test.tsx`, which I had not looked at when I wrote that.


wlsdks:

PR #1777, opened 2026-09-19
fix(map): a tier name that lost its plane goes to the corner

wlsdks:
## Summary

Strata names its four planes on a rail down the right edge. The rail's whole claim is that a row's height answers "which ring is this name for" without the reader counting rings. The rail begins under the last utility tile, so a plane projected above that tile is out of reach.

Measured at 1512×982 with the sample folder expanded:

| tier | rail row | its plane |
|---|---|---|
| project | 330 | 124 |
| domain | 350 | 316 |
| capability | 531 | 507 |
| element | 704 | 614 |

The project's name stood **205 px** from the only project node, ten of its own rows, and dragged the domain row down until the two touched at exactly one row's spacing while the pair below them sat about 175 px apart.

The rail now hands over to the corner stack when a row loses its plane. The corner gives up per-plane alignment out loud, which is honest, where a misaligned rail is a rail that lies. The rail is not removed: it is used wherever it can keep its promise.

Two attributes changed so the DOM stops misreporting: `data-tier-legend-placement` now names the shape actually drawn rather than the one asked for, and `data-tier-legend-offset` publishes the worst distance between a row and its plane while the rail is drawn, which is what the new gate reads.

## The existing size fixture moves, with its reason

`map-3d-strata-drawing.spec.ts` pinned 1512×982 to `rail`. A free column is all that fixture asked about, and a free column is not a reachable one — it never checked that the rows reached their planes. That entry becomes `corner` with the reason written beside it, and the rest of that gate, which is that the names land on nothing, still runs unchanged at both sizes.

## Routing

`pnpm design:route -- --change=layout --change=topology-encoding` returns `directions=no · council=no · seats=none`, sequence `build>proof`. No council was required.

## Test plan

- New `tests/e2e/map-strata-tier-scale.spec.ts` pins the rule rather than the outcome: while the rail is the shape drawn, no row may be further than its own height from the plane it names; the corner branch instead requires evenly spaced rows, which the clamped rail could never produce. RED without the fix, reporting 205 px against a 20 px row.
- Four new cases in `tier-legend-rows.test.ts` for the pure predicate, including that the tolerance is the row's own height rather than a number of its own.
- `pnpm exec eslint --max-warnings 0` on all three changed sources
- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run src/widgets/ontology-map/model/tier-legend-rows.test.ts` — 13 passed
- `pnpm source:language` — ratchets current
- `pnpm knip` — clean
- `pnpm test:contracts` — 2,729 passed
- `map-3d-strata-drawing` at both sizes plus the new spec — 7 passed

## Known, not caused here

Sampling the legend twice a second for 10 seconds, 9 of 20 samples catch it mid-remount with no band measured, which turns the rim names back on for that frame. That is the folder-watch rebuild measured in #1772, which remounts the map every few seconds; it happens on `main` with or without this change.


wlsdks:
`map-3d-strata-drawing.spec.ts:456` is red on every pull request that builds the export fresh, and main's green does not cover it.

Measured from two of my own pull requests (#1734, #1736), neither of which touches `src/widgets/ontology-map` or this spec:

```
Strata 1512x982 — the tier legend names four planes without landing on anything
  Expected: "corner"
  Received: "rail"
```

Identical on both, and on both retries inside each run — six attempts across two independent CI runs, same assertion, same width. The 1040x720 case passes every time.

**Why main looks green.** The two runs differ in one job:

| | Build browser artifact | Playwright (chromium 3/3) |
|---|---|---|
| main `a975e80ae` | **skipped** — cached export reused | success |
| #1734 / #1736 | **success** — built fresh | **failure** |

So main's post-merge run tested a cached export, not a fresh build of current main. A pull request builds fresh, and that is where the assertion flips.

**It does not reproduce locally**, in either run mode, on a tree that is current main plus four unrelated files: `pnpm exec playwright test` under dev passes, and so does `pnpm build && PLAYWRIGHT_STATIC=1 pnpm exec playwright test` after a full rebuild. Both widths pass locally. So the placement depends on something the runner has and a developer machine does not — font metrics are the obvious candidate, since the choice between the rail and the corner is a question of whether four names fit, and locally they do not fit (corner) while on the runner they do (rail).

Worth noting that `4b880ba43` (#1702, "a project is drawn by its locale display name on every screen") landed immediately after this one, and it changes which strings the map draws. The two were never tested together on a fresh build.

I am not touching the spec or the placement: which way a tier name that lost its plane should go is your decision, and I cannot verify a fix against a failure I cannot reproduce. Flagging it because it blocks landing for anything based on this main.

PR #1779, opened 2026-09-19
fix(pr-land): a read that did not happen is never rendered as an answer

wlsdks:
## Summary

Tonight twenty-six landers shared one GitHub account, the REST quota reached `0/5000`, and `pr-land` began stating facts it had failed to fetch. Four reads in this script tolerated failure by returning the same value they return for a real answer, so a refused request and a genuine result were indistinguishable.

**The lock.** `pnpm pr:queue` printed `nothing is landing` while #1728 held the lock and was refreshing it every minute — `readLockPayload` collapsed a 403 into the same `null` that means "there is no lock". The comment above `classifyLock` already said an unreadable lock must not read as an open door; the guard sat one layer above the read that actually fails.

**The lock, worse.** When the ref read is served and the blob read is refused, a live lock arrives as `{}`, classifies as `unreadable`, routes to `take-stale-lock`, and is force-PATCHed over. Two landers then both believe they hold it and both merge into main. This is not theoretical: another session logged `taking over a lock nobody refreshed: an unreadable lock ({})` at 08:06 while the quota was pinned.

**The gate.** `readRequiredContexts` turned a refused request into an empty list, and an empty list makes the lander print `main declares no required status checks; refusing to land without a gate`. That accuses the repository of having no protection, and the obvious way to act on the accusation is to weaken branch protection until a landing proceeds.

**The base.** `readBehindBy` turned a refused request into `0`, which reads as "main has not moved", so the landing skipped pouring main in and spent its one CI run on a base nothing had checked — the inverse of that step's purpose.

## What changes

A read that did not happen is now its own state everywhere:

- `readOutcome` tells a genuine `404` from a request that never arrived, and `describeReadFailure` names the rate limit rather than leaving an empty brace for the next session to interpret an hour later.
- `classifyLock` gains `unknown`, which `decideNext` routes to **wait**. A blob that genuinely arrives empty still classifies as `unreadable` and still recovers a stale lock — that arm is deliberately preserved, because a careless version of this fix turns a wedged lock into a permanent one.
- `--release` refuses to delete a lock it could not read, instead of force-deleting one somebody is holding.
- `protectionFrom` keeps `404` as a real answer (main may genuinely have no rule) but never lets a failed read make the accusation.
- `distanceFrom` treats both a failed read and a bodyless comparison as unknown, and the loop waits a poll rather than landing on a guess.

Asymmetry worth noticing in review: a `404` is a real answer for the lock ref and for branch protection, and no answer at all for a comparison. Each is commented where it is decided.

## Test plan

- `pnpm test:pr:land` — 71 pass (66 before).
- `pnpm checks:changed -- --run` — 3 passed (`source:language`, `test:pr:land`, `knip`).
- **Deliberate RED, each reverting one guard and reddening only its own case:**
  | reverted | fails |
  |---|---|
  | `classifyLock`'s `unknown` state | 2 cases |
  | `decideNext`'s wait on `unknown` | 1 case |
  | `readOutcome`'s 404 discrimination | 1 case |
  | `unreadable` collapsed into `unknown` (the stale-recovery arm) | 2 cases |
  | `protectionFrom` answering `[]` on failure | 1 case |
  | `distanceFrom` answering `0` on failure | 1 case |

  GREEN at 71 after each restore.

This cannot help a landing already in flight — a running lander holds its own copy of the script — so it is for the next outage.


wlsdks:
**Known remaining gap, measured by another session tonight rather than by me.**

This change makes the lock read, the branch-protection read and the `main...head` comparison say when they could not be served. It does **not** cover the required-checks read, and that is the one a lander actually dies on. A lander that gave up printed:

```
gave up after 180 minutes; the lock is released and nothing merged
```

and the eight contexts it had been waiting on all passed a short time later. So the 180-minute ceiling can spend a full CI round for nothing when the *read* failed rather than the run — the same defect class as the four fixed here, in the one place that costs a whole round.

The fix has the same shape: a give-up should distinguish "the checks are still running" from "I could not read the checks", and only the first deserves the wait. Left out of this pull request deliberately, so the change stays one track and lands on its probes.

wlsdks:

PR #1781, opened 2026-09-19
fix(acp): the app's login copy is filed where Claude Code looks

wlsdks:
## Summary

Loop round 55 (2026-09-20), priority ①: what the Library does with a connected agent must actually work. Since yesterday the installed app could not start a single agent turn — every press logged `acp session acp-N exited with code Some(0)` and the conversation showed *「이 앱 몫으로 남은 옛 로그인이 가로막고 있어요」*, whose retry loops. Rounds 52–54 measured around it and got it wrong twice (a duplicate keychain item, then a closed stdin). This names it.

**A keychain item is named by service *and* account.** The mirror wrote the terminal's login under the literal account `claude`; Claude Code reads and writes the macOS login name. Two items therefore lived under the app folder's service, `security -s <service> -w` answered with whichever it reached first, and the child read only its own — so the copy the app had been writing for weeks **had never once been read**.

What the child did read was the item it had migrated out of the linked `.credentials.json`. On this machine that file is a leftover the terminal stopped writing: its access token had expired (`expiresAt: 0` after migration) and its refresh token was spent. Claude Code migrates such a file into the keychain and deletes it, so every start relinked the fossil and every start re-poisoned the folder. That is why the screen said the login had expired while the very same bytes answered in the terminal.

Three changes, all in `src-tauri/src/acp.rs`:

- `CredentialCarrier` carries the `acct` it was read under, `parse_keychain_account` reads it, and `mirror_account` takes the account **from the carrier** instead of naming a literal. A file carrier has no account, so the home folder's own name stands in.
- A successful mirror takes the symlink down, so the migration cannot write over it again. Only a symlink is removed; a real file in the app folder is never the app's to delete. The link and the existing `clear_shadowing_credentials` path are untouched when nothing was mirrored.
- The mirror stops at a synthetic home. Every test handed one in and reached the real login keychain, leaving one item behind per run — the owner's machine holds 3,104 `Claude Code-credentials-*` items.

## Evidence

- Minimal reproduction without the app, one ACP stdio client against `@agentclientprotocol/claude-agent-acp@0.70.0`: `session/prompt` answers normally with no `CLAUDE_CONFIG_DIR`, and returns `Failed to authenticate: OAuth session expired and could not be refreshed` with the app's folder. The one difference is that variable.
- Decisive step: deleting the item filed under the login name and leaving only the app's `claude` copy turned the error into `Authentication required` — the mirror is not read.
- Installed app rebuilt from this branch, trial vault `acp-trial-small`: press 서식 맞는지 확인 → `acp login mirrored from Claude Code-credentials into the app-scoped keychain item for jinan`, no exit line, the agent reads three wiki pages and returns a report with `superseded 1` and a fix door. `wiki/_log.md` gained `## [2026-09-19T23:41:09.607Z] lint | disagreement 0 · superseded 1 · missing-link 0 · name-without-page 0 | agent:claude-acp`. Shot: `.claude/shots-2026-09-20/300-acp-turn-works-after-account-fix.png`.
- Rust: `the_mirror_is_filed_under_the_account_claude_code_itself_reads` (verbatim `security` output → account; the account comes from the evidence; a file carrier falls back to the home folder's name). `cargo test --lib` 324 pass, `cargo fmt --check` and `cargo clippy` clean for this file.

## Test plan

- [x] `cargo test --lib`, `cargo fmt --check`, `cargo clippy`
- [x] Installed app: a real agent turn end to end, proven on disk
- [ ] `pnpm checks:changed -- --run`


wlsdks:

PR #1782, opened 2026-09-19
fix(map): names stop being painted under the open inspector

wlsdks:
## Summary

The label cull asks a safe rect whether a name has room, and that rect came from `--map-safe-inset-right`, a static 120 px. The inspector that opens on a selection is 352 px wide from x=1128, so the cull believed there was clear canvas where a panel was sitting.

Measured at 1512×982 on the sample folder, clicking each of the five concepts the first frame offers:

| | before | after |
|---|---|---|
| Selections that painted a name mostly under the panel | 2 of 5 | 0 of 5 |
| Names lost that way | 주문, twice | none |

The name could not be read, and it still spent one of the map's limited label slots.

The camera has taken the larger of token and measurement since 2026-08-10. This gives the names the same truth, left and right only, for the reason that note already gives: the top inset is the tool lane and the bottom one is a label reservation, so neither is something covering the canvas. The frame reads its tokens every frame and measuring walks elements, so the draw reads a ref the camera fills at the moments a panel opens or closes, rather than measuring per frame.

## Test plan

- New `tests/e2e/map-no-label-under-the-panel.spec.ts` states the rule rather than the case: no label may be painted mostly under an open panel, for whichever panel a selection opens. RED without the change, naming both selections and the name each lost.
- `pnpm exec eslint --max-warnings 0` on both changed files
- `pnpm exec tsc --noEmit`
- `pnpm source:language` — ratchets current
- `pnpm knip` — clean
- Escalated because this narrows the label cull: every label-related map spec — `map-ego-dim-labels`, `map-label-collision`, `map-rail-labels`, `map-spotlight-label-clearance` plus the new one — 8 passed.


wlsdks:
Heads-up for whoever owns the map loop: `map-3d-strata-drawing.spec.ts:456` fails on every pull request that builds the export fresh, and it blocks landing for anything based on current main. Full evidence is on #1777, which introduced the expectation and has since merged — leaving it here too because that thread is closed and this branch is live.

The short version, with the numbers:

```
tierLegendPlacement:  freeWidth - 56 >= freeHeight * 1.08  ?  "rail" : "corner"
spec at 1512x982 expects "corner";  CI gets "rail"  (1040x720 passes)
```

At 1512x982 the boundary sits at a free width of about 1115 px. With the INDEX panel open the free width is about 1148 — **rail by roughly 31 px**. So "corner" at this size is within about 31 px of flipping, and whatever narrows the box that far locally is not happening on the runner.

Reproduced directly against the built static export at 1512x982: `data-tier-legend-placement` reads **rail**. Two of my pull requests (#1734, #1736), neither touching `src/widgets/ontology-map`, fail it identically — six attempts across two independent CI runs, same assertion, same width.

Why main looks green: its post-merge run has `Build browser artifact` **skipped** (cached export reused) while a pull request builds fresh. Main has never tested a fresh build of its own head against this lane.

Also worth a look: `4b880ba43` (#1702, locale display names on every screen) landed immediately after #1777 and changes which strings the map draws. The two were never tested together on a fresh build.

I am not touching the spec or the placement — which way a tier name that lost its plane should go is the map loop's decision, and the 31 px boundary is the part that needs deciding, not just the assertion.

wlsdks:

PR #1783, opened 2026-09-20
fix(library): a rewritten answer is not a drifted one

wlsdks:
## Summary

Loop round 56 (2026-09-20), priority ①. With agent turns working again (#1781), I walked the whole Library loop in the installed app on the trial vault: lint → the finding's fix door → a real turn with three allow-once presses → both pages rewritten → `_log.md` fix line. The loop holds. This is the one thing it got wrong.

**A rewritten answer is not a drifted one.** The retained answer's evidence card judges drift from `answer_source_observations`, which is immutable by design — it records the bytes the answer read. Once a source moves, that comparison says *changed* for good, including in the second right after a fix turn rewrote the page from the new bytes and kept the old figure under the new one. So the reader offered *에이전트에 갱신 초안 요청* for work that had just been done, while the wiki row beside it said *확인된 원문 버전이 그대로입니다*. Two surfaces, one file, one moment, opposite claims.

The row is right, and it is right because it reads `source_hash` — which every writer fills with the sha256 of the bytes it read (`compile-brief.ts`), so it is the evidence for *this page is already written from this version*. `answerObservation` now reads it too: a source whose bytes moved on, on a page whose own `source_hash` names the bytes on disk today, is **`rewritten`**.

- A sixth state with its own sentence in both locales rather than folding into `unchanged`, because *no change since source observation* would be a fresh untruth — the source did change.
- In the question list it takes the check the row grammar already gives a state with nothing to act on, not the amber badge that asks for work.
- Under the evidence disclosure it names which sources, beside *바뀐 원문* and *없는 원문*.
- The refresh button stays offered. A person may still want a fresh draft; the headline just no longer claims the page is behind.

## Evidence

- Installed app, rebuilt with this change, trial vault `acp-trial-small`, the exact page the fix turn had just rewritten: the card reads *인용한 원문이 바뀌었고, 이 답은 그 판본으로 다시 쓰였습니다* and the disclosure reads *다시 쓴 원문: sources/refund-policy.md*. Before: `.claude/shots-2026-09-20/301-fix-door-turn-done.png`. After: `.claude/shots-2026-09-20/302-answer-rewritten-not-drifted.png`.
- Unit: `answer-revision.test.ts` — *reads a page rewritten from the new bytes as caught up, not as drifted*, pinning all four: rewritten, a page still recording the observed hash (`changed`), `source_hash: unmeasured` which a filed answer starts with (`changed`), and a source that never moved (`unchanged`). 19 pass with the two component suites.
- `pnpm test:i18n:messages` 17 pass; `pnpm checks:changed -- --run` 10 passed.

## Test plan

- [x] `pnpm checks:changed -- --run`
- [x] Installed app: the page a fix turn rewrote, read back


wlsdks:

PR #1784, opened 2026-09-20
design(acp): the write condition is a time a person can check, not thirteen digits

wlsdks:
## Summary

In the permission card's Details tab, the row reporting the write guard renders its condition raw:

```
expected_mtime   1727000000000
```

That number *is* the condition — the file is written only if it has not changed since that moment — and thirteen digits answer nothing the reader came to the row to ask. The question the row exists for is "is this the version I am looking at?", and a date answers it. It now renders through the locale's own date and time format, in both catalogues.

The exact value is led, not replaced: the epoch stays on the row's `title`, so an agent or a terminal reader still has it to the millisecond. Only a `_mtime` key carrying a finite number is read as a time; every other write condition keeps the value it arrived with.

## A defect this branch found in the local gate

The card's new dependency on `useFormatter` broke **25 cases** in `AcpChatPanel.test.tsx`, which mocks `next-intl` by hand and had no such export — so the card threw on render and every one of those cases failed on an empty `<body>` rather than on what it was asserting.

`pnpm checks:changed -- --run` passed six lanes and never selected that suite. The two tools answer different questions: `checks:changed` maps changed paths to lanes by rule, while the pre-push `unit` lane runs `vitest --changed` from the merge base and walks the module graph, so it finds a test that imports a component you touched even when their filenames share nothing. Here the changed file is `AcpPermissionCard.tsx` and the suite that renders it is named after its sibling panel, so no path rule connects them.

Worth knowing beyond this branch: a suite that renders a component through a hand-written library mock will fail this way — silently, on a blank body — every time that component grows a dependency.

## Test plan

- `pnpm exec vitest run src/widgets/acp-chat-panel/ui/AcpChatPanel.test.tsx` — **133 passed** (25 failing before the mock fix; the same suite passes 133 on today's main, confirming the failures were this branch's).
- `pnpm checks:changed -- --run` — 6 passed.
- Pre-push lanes — typecheck, source_language, comment_refs, decisions, unit — all pass in 55s.

Opened late: this round's work was committed hours ago but never pushed, so the pull request is catching up with the branch, with today's main merged in.


wlsdks:

PR #1786, opened 2026-09-20
chore(acp): the registry follows upstream, and the pins say what was read

wlsdks:
## Summary

`pnpm acp:registry:check` was red on **every** branch that touches the ACP launch path, so no such pull request could reach CI. #1781 proved it: it took the landing lock, ran the focused lanes on the merged source, and stopped there. Fifteen listed agents had moved upstream, two of them runtimes this app actually runs — `claude-agent-acp` 0.77.0 → 0.79.0 and `codex-acp` 1.11.0 → 1.12.0.

A version bump alone is not evidence, and three contract gates exist to say so. Each pinned surface was re-read from the published tarball and compared byte for byte with what the repository transcribed:

| Pinned surface | Gate | Result |
|---|---|---|
| claude `dist/permissions/options/` | `acp-permission-option-choice` | byte-identical to 0.77.0; every recorded digest unchanged |
| claude `dist/session-mode.js` | `acp-gate-safe-modes` | byte-identical, same SHA-256 |
| codex `AgentMode` (111 lines) | `acp-runtime-gate` | byte-identical; only bundled `@openai/codex` moved, 0.153.4 → 0.154.0 |

`optionId` still appears in no `dist/**/*.js` outside `dist/permissions/`, so nothing else builds an option array.

One file did change under `dist/permissions/`: `presentation.js` stops compacting a Bash or PowerShell title, because compacting whitespace in executable input changes quoted arguments and comment boundaries. That is the request's wording, not its option set or its modes, so no transcription needed rewriting — and the note in each gate now says that, so the next reader does not have to open the tarball to learn it.

## Evidence

- Tarballs fetched from the registry directly (`npm pack` is refused by the publish guard) and compared with `diff -rq`.
- The three gates: 73 tests pass.
- `pnpm checks:changed -- --run`: 6 passed, including `acp:registry:check`, `desktop:check` and the web-surface smoke.

## Test plan

- [x] `pnpm checks:changed -- --run`


wlsdks:

PR #1791, opened 2026-09-20
fix(ci): a measurement runs where the number means something

wlsdks:
## Summary

`node-name-match.perf.test.ts` reddened three times — **6.73, 9.20, 9.20** against a bar of 10 — on branches that never touched the matcher it guards, and never once near the **2.7–5.3** its real defect produces. Reported twice in the same place by another session rather than re-run, which is what made it diagnosable.

**The gate's own doc-block was wrong, and I wrote it.** It claimed the ratio was load-proof because numerator and denominator ride the same machine and the same concurrent load. They do share the machine; they do not share the **allocation**. The cached side holds a per-node index and pays the collector under memory pressure; the naive side walks strings it throws away. So contention pushes the ratio down from above while barely moving the bottom, and the gate reddens on a fast path that works. Vitest shards by file, so the verdict depended on which other files happened to land in the same shard — a lottery, not a gate.

## What changes

`.claude/rules/testing.md` already said performance lanes run *"where the number means something: CI on a quiet runner, not the pre-push hook, which deliberately saturates the machine."* A **sharded** lane is not a quiet runner either. This gives that rule somewhere to live:

- **`vitest.config.ts`** — a `perf` project owning `**/*.perf.test.*` with `fileParallelism: false`, so no sibling worker runs on the machine while a measurement is taken. The `jsdom` project excludes them.
- **`run-ci-lane.mjs`** — every sharded sweep names `--project=contract-node --project=jsdom`, and `pnpm test:perf` runs the measurements once, alone, on the shard that already carries whole-graph work.
- **`.claude/rules/testing.md`** — records that a sharded lane is not quiet, with the measured cost of getting it wrong.

**All six measurement files** were on this lottery, not just the one that hit it: `build-local-manifest`, `derive-ontology-from-vault`, `duplicate-pairs`, `library-force-simulation`, `library-islands-layout`, `node-name-match`. `duplicate-pairs` is the model the rules cite, and it was sitting on the same risk without having tripped.

**The threshold is unchanged on purpose.** Re-deriving it under load would encode one afternoon's contention and overlap the healthy and defective bands, leaving the gate nothing to separate. The corrected doc-block says so, and says that if it reddens again the lane moves, not the number.

## Test plan

- `pnpm checks:changed -- --run` — **18 lanes pass**.
- `pnpm test:perf` — 6 files, 10 tests, 14.2s.
- **Deliberate RED, both directions:**
  - make `nodeNameIndex` ignore its `WeakMap` → the gate still catches it after the move: `cached 41.2ms · rebuild-every-tier 143.9ms · ratio 3.5x`, inside the documented defective band. Restored → green.
  - drop the project exclusion from the sweeps → the new contract case `the measurement files never ride a shard, and run once on their own` goes red. Restored → 69 pass.
- Verified by listing rather than by reading the config: `vitest list --project=contract-node --project=jsdom` selects **0** perf files; `--project=perf` selects **6**.


wlsdks:

